PR #14, opened 2016-02-29
Travis ci integration and auto self unittest

Shoelace:
Ok.. so with a bit of finagling i have been able to setup a travis-ci.org build setup.

when configured travis will 
* download oracle xe 11gr2
* install oracle
* create utp user
* install utPLSQL under utp user
* run self unit tests
 * tests all installed packages that begin UT_.  (the current code has 2 UT_UTOUTPUT and UT_UTREPORT )

the build will 'fail' if any unit test has a failure. 

this is mostly proof of concept.. the travis-ci setup needs oracle technet credentials

if pulled to master the readme.md needs to be updated with teh correct url for eth badge


Shoelace:
i'll add that im happy to import request to go thru travis..
or i can give write access to my version if anyone want to try it out..

Shoelace:
oh.. and sample output for thsi commit
https://travis-ci.org/Shoelace/utPLSQL/builds/112625117

jgebal:
This viscerally great.
I've looked at travis and it seems brilliant for collaboration.
Continuously checks project health including pending pull requests. Really,
really cool.
Would be great to have this setup for main project as well as v3.

PaulWalkerUK:
Do you want me to merge this to the "master" branch now, or would it be better on the newly created "version3" branch?

I haven't looked at how Travis does an Oracle database - I assume there are no licence issues to worry about?

jgebal:
No issues the way it's used now. We do not redistribute the binaries. This
approach is slower for CI but this is ok. CI should be a safety and health
check. This one seems perfect. The trouble is 12c as there is no
lightweight version to work on. I didn't use Oracle cloud solutions and
don't know if that could be an option.

jgebal:
I think keeping CI for her son three only is a good choice.

Shoelace:
yeah.. it was more proof of concept..  i'll close this request as is.. and re-apply to whatever comes next..

Issue #23, opened 2016-07-18
Compatibility with 11.2


Pazus:
Hi all.

I've forked the version3 branch and unfortunately i've figured out that it is not compatible with 11.2 due to the use of record types in Native Dynamic SQL  statements with only appear to be included in 12c. It might slightly change the idea to store all types in ut_types package.

Do you think records with colelctions should be replaced with object types?

Also I think definition of _dbms_id_, _dbms_quoted_id_ types should be made with conditional compilation to support both 11g and 12c.

Shoelace:
yes i agree about the conditional comp of dbms_id and dbms_quoted_id.  i had planned to do that i just hadnt gotten to it.

you seem to be rigth about teh execute immediate restriction.
compare
https://docs.oracle.com/cloud/latest/db112/LNPLS/dynamic.htm
to
https://docs.oracle.com/database/121/LNPLS/dynamic.htm

object types are nice as it allow you to add methods to the objects (such as validate() or toString() ) rather then using helper functions.  i dont think there are any licencing issues regarding them either and are valid back to at least 10g..  but i think we decided we were only going to support 11 and up anyway.

i'll try and dig out my 11.2 version and see if i can come up with another solution.

thanks for the heads up..

Pazus:
If you are not against the object type approach I can prepare a pull request.

Pazus:
The problem is that one can't use conditional compilation for object type properties, so untill 11g DB version is supported dbms_quoted_id is not usable.

Shoelace:
well that was just a nicety.. there is no real reason to not use varchar2(32)  i just found dbms_id as i was rumaging and thought it would be cool.

Pazus:
I have found out that the main idea of your approach through records and NDS doesn't fit 11g restrictions. As there is no straight forward way to pass record type variables to NDS it seems that object based design might be the only way to go. 
Also it has other benifits: we can define abstract reporter and one can enrich reporting solutions by inheriting abstract object type with custom implementation, this objects would be called without any dynamics by methods from the abstracct class. OOP Composite pattern, mentioned in #17, where suites can be grouped into larger suite will be difficult to implement through records.
All this encourage me to make a object type based implementation of the v3 according to all the restrictions and compatibilities discussed. I'll let you all know when I finish me draft but beforу that I would like to hear you opinion of such design because it has drowbacks like difficulties with altering types that form referencing other types in the future.

Shoelace:
[...]

PR #24, opened 2016-07-21
Object types design

Pazus:
Implementation was discussed in #23.
This was tested on 11.2.0.4 database.

This is a complete rewrite based on an object model to use features like inheritence, compositor pattern to create tree-like suites composed of suites, custom reporters and decorators for plugable enrichments based on inheritance of abstract class and interface.

A lot of work has to be done on enchancing assert capabilities, suite manager to automaticaly generate suits and test based on test-package methods. Though this might be a base of the framework.

As there is no naming convention defined for the project I tried to stay with the naming style used befor in v3 branch. But I thick naming convention has to be clarified in the future.


Shoelace:
i dont see anythign wrong with this..  most of my 'suggesions' are more just additional developemnt.

the only thing here that i see different from the current (that i had worked on) is teh use of do_resolve and the setting `owner_name         => coalesce(trim(a_owner_name)                                                                          ,sys_context('userenv', 'current_schema'))`  in ut_test.tyb

in the previous version this was left blank if not specified and backfilled by do_resolve.  i did this becuase running `package.procedure;` is not the same as `user.package.procedure` if there is a synonym in place.

i assumed it would be best to leave it to  dbms_utility.name_resolve to fill the owner so the 'VALID' check works correctly.  i think the catch is it will need to be authid current_user to work..  but a lot of stuff will need to be changed on that front.

oh.. and i dont think its in teh proposed standards yet but i think most of us would prefer varible names and comments in english.  which include the git commit comments.

Pazus:
If you like I can user sql formatter first and put settings file along other files of the project

jgebal:
Putting code through formatter sounds like a good idea

Pazus:
@Shoelace I use dbms_sql for dynamic SQL invocation as dbms_sql uses Current user rights to prevent from unauthorized access ([11.1 docs](https://docs.oracle.com/cd/B28359_01/appdev.111/b28419/d_sql.htm#i1027587)). There is not need for authid current_user with such approach.

Shoelace:
I wasnt aware of the feature of dbms_sql. But it does make sense when you think about it. So cool. 1 less hassle for now. :D


_____________________________

Pazus:
@Shoelace Sorry for russian commit comments in between I tought pull commit translates into single commit in the main project. Whatever I combined all temp commits into one so it's not a problem any more.

Also I have returned do_resolve behavior to the one you had and mad some additional changes

jgebal:
I've ran all the examples on 11g XE.
All running and passing.
This is a major rework so I'd suggest merging this PR, so that others can contribute to the new codebase.
@Pazus , did you apply the formatting provided by @Shoelace ?

examples/RunExampleComplexSuiteWithDBMSOutputReporter.sql
```diff
@@ -0,0 +1,45 @@
+--Shows how to create a test suite with the default reporter which is dbms_output
+--No tables are used for this.   
+--Suite Management packages are when developed will make this easier.
+Clear Screen
+Set Serveroutput On Size Unlimited
```
jgebal: I would change the example scripts to make them self contained by:
- calling script to create example ut package(s) before execution of script
```sql
@ut_exampletest.pks
@ut_exampletest.pkb
@ut_exampletest2.pks
@ut_exampletest2.pkb
```
- drop the package after the script is finished (optionally)

Pazus:
We will do that shortly, I'll create new pull request.

Pazus:
I have merged PR from @jgebal which has some minor improvements. I hope now this PR is ready to be merged also.

rlove:
I like the design changes.

PR #25, opened 2016-08-01
Object types design

Pazus:
Class model rework based on discussion with @jgebal

All the hierarchy now is formed from ut_object class, asserts, tests and suite are homogeneous by composite pattern.

Lots of new tests/examples added
Composite reporter class added to use multiple reporters simultaneously
reporters now accept ut_object (ut_suite, ut_test, ut_assert) as an input and are statefull

Some classes where dropped.


rlove:
I see no problems with this pull request.

PR #31, opened 2016-08-15
Composite reporter is now working as expected.

jgebal:
The reporter was missing assignment in the constructor.


rlove:
Won't compile after accepting variable rename #32 pull request.

jgebal:
Fixed

rlove:
Accepted.

PR #32, opened 2016-08-15
Local variables, global variables and arguments now to match naming c…

jgebal:
…onventions.

Existing code was refactored.


rlove:
Code looks good.

PR #34, opened 2016-08-16
Preparation for the annotation configuration approach and suite setup/teardown

Pazus:
Fixed custom reporter in the example after naming convention fixes
Moved Add item to the ut_composite_object type
Added object_name attribute to the ut_test_object type to save the procedure/package name for test/suite, for the suites primary to construct "suitepath" in the annotation approach
Added search procedure to find an index of the item with defined object_name in the array of ut_composite_object
Added separate setup and teardown procedures to the suite which run at the beginning of the whole suite and at the end. Currently invalidation of the all nested object is not developed

All the examples and tests do pass.

### DIFF
--- a/examples/RunExampleComplexSuiteWithCustomDBMSOutputReporter.sql
+++ b/examples/RunExampleComplexSuiteWithCustomDBMSOutputReporter.sql
@@ -43,6 +43,6 @@ begin
 end;
 /
 
---drop type ut_custom_reporter;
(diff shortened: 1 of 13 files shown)


Pazus:
Made the default empty implementation for ut_suite_reporter methods so it will be easier to add new in the future without invalidating all it's subtypes that doesn't have required implementations of the new methods

PR #40, opened 2016-08-19
Travis cache

Shoelace:
okay.. so here is a pull request that adds travis directory cache functionality.

the oracle file is now downloaded to $ORACLE_ZIP_DIR  which defaults to $HOME/.cache  or can be set using travis variables.

the rest of the details are from here https://github.com/cbandy/travis-oracle/pull/17

i have also moved NaturalDocs to the tools directory  (which is ignored when github does its repo languages stats)


Shoelace:
i should say the point of this is so pull request will work with travis.. 

this one will fail but future ones should work  (but untested as of now)

Shoelace:
only 3 of the 123 files have been editied.. the other is the moved directory.

PR #42, opened 2016-08-23
Updated install script.It now fails on invalid sources after installation

jgebal:


rlove:
Been thinking about this, and what if someone install into an existing schema that already contains invalid objects.

Maybe we should change the query to something like
```
 select count(1)
    into l_cnt
    from user_errors where name like 'UT%';
```

jgebal:
Good idea. I'll fix it

jgebal:
I did more thinking around the install validation.
First of all, we should keep the 'reporting' select in sync with the 'validation' select.

I'm not sure if we should really take into consideration the fact that the schema where utplsql is getting installed is dirty (contains invalid objects).

The ideal solution would be to only lookup exceptions for objects that are part of utplsql installation.

PR #44, opened 2016-08-25
Annotation parser performance improvements

Pazus:
- added some comments
- switched to $$ut_trace compilation flag
- refactored config_package procedure specification
- added huge package for parsing tests
- added example of parsing huge package spec
- switched to all_source/dba_source views depending on current grants
- redefined objects exclusion list for config_schema procedure


Pazus:
Don't understand why new examples didn't execute...

PR #54, opened 2016-10-01
Cleanup and separation of annotation processing.

jgebal:
Now annotations are separated from DB metadata processing.

### DIFF
--- a/source/ut_metadata.pkb
+++ b/source/ut_metadata.pkb
@@ -1,8 +1,10 @@
 create or replace package body ut_metadata as
 
+  ------------------------------
+  --private definitions
   g_source_view varchar2(32);
 
-  function get_source_cursor(a_source_view varchar2, a_owner varchar2, a_object varchar2) return sys_refcursor is
+  function get_package_spec_source_cursor(a_source_view varchar2, a_owner varchar2 := null, a_object varchar2 := null) return sys_refcursor is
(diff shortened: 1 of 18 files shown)


Pazus:
Lets change `ut_annotations.parse_package_annotations` to accept clob as input parameter so it is more testable so the source getter should be moved to the `config_package` procedure.

Pazus:
Does the cursor in the `get_package_spec_source` has to be explicitly closed on exception?

jgebal:
We don't need to close it as it failed to open.
Trying to close it raises:
```
ORA-01001: invalid cursor
ORA-06512: at "UT3.UT_METADATA", line 144
ORA-00942: table or view does not exist
```

source/ut_metadata.pkb
```diff
@@ -102,319 +111,37 @@ create or replace package body ut_metadata as
...
     g_source_view := 'dba_source';
-    close l_cur_temp;
-
-  exception
-    when others then
-      g_source_view := 'all_source';
+   exception
+     when l_object_does_not_exist then
+       g_source_view := 'all_source';
   end;
```
Pazus: Could you please steal "lazy" implementation from my PR to your branch, because sys_refcursor causes a bug (ORA-600) on our setup. Don't merge other staff.

jgebal: Not sure why the merge of PR didn't simply accept your changes.
I'll do a pure merge without a PR to update my branch.

PR #69, opened 2016-10-07
Overview, v2 to v3,  Status Shields

rlove:
Update the readme to include status shields and an initial comparison of v2 to v3 features.

To see the changes rendered...  Click on "Files changed", and then "View" on the file.


Pazus:
Should we mention that now nested suites are supported?
Should we mention framework operated transaction control?
Also in v3 we can use multiple reporters simultaniously

jgebal:
I think all of above is worth mentioning.

rlove:
@Pazus and @jgebal - I added all three.

Issue #88, opened 2016-10-18
Rework naming and folder structures


jgebal:
As project grows it gets more and more important to have it well organized and elements properly named.
This story is to address this

rlove:
I have cleaned up and removed `tools` and `docsource` in pull request #94 

This really leaves `build` directory which really seems out of place.

PR #89, opened 2016-10-19
Restructured directories and renamed files.

jgebal:
This is addressing issue #88


Shoelace:
can you explain your logic on your type renaming..

ie. assertion  , expectation and matcher etc..

jgebal:
Hi @Shoelace 
expectation = assertion (synonym). I've decided to move away from assertion and lean towards expectation, as we now use syntax: ut.expect...

Matchers are "operators" used to perform comparison of actual to expected.
Base object type for matchers was badly named ut_expectation before.
This pull request addresses two things.
* naming in of object types for expectations (aka asserts)
* directory structure of source code (partly)
As the project grows it seems more valid to group code by functional areas rather than object types.
This is why I've created subdirectories in `source` directory:
* **core** - for the code of the ut engine, including abstract and base object types for extensions
* **reporters** - for specific reporters
* **expectations** - for all the expectations and matchers

@rlove, @Shoelace, @Pazus, what do you think about this?

Pazus:
Agree

PR #93, opened 2016-10-20
Feature/warning cleanup

Shoelace:
this PR is desigend to cleanup the code in regards to teh PLSQL compiler warnings.

for each commit (or pair of commits) i tried to perform 1 type of cleanup in case some to allow easy cherry-picking later if desired.


Shoelace:
crap.. i've just notice i've build this on top of #92 feature/naming_check 
so that should be sorted first

PR #99, opened 2016-11-12
Teamcity reporter

Pazus:
Serveroutput based reporter for Teamcity CI via "service messages".


source/install.sql
```diff
@@ -133,10 +137,12 @@ whenever oserror exit failure rollback
 
 
 prompt Validating installation
+-- erors only. ignore warnings
 select * from user_errors 
 where name not like 'BIN$%'  --not recycled
 and (name like 'UT%' or name in ('BE_FALSE','BE_LIKE','BE_NOT_NULL','BE_NULL','BE_TRUE','EQUAL','MATCH')) -- utplsql objects
-and attribute = 'ERROR'; -- erors only. ignore warnings
+and attribute = 'ERROR'
+/
```
jgebal: Is this change making the SQL display what is failing?
I've noticed that the current install script doesn't show errors, even if they exist.

Pazus:
This change is made as current script fails in PL/SQL developer command line, don't know why but with slash it works and results are displayed.

jgebal:
Looks good - merging.

PR #102, opened 2016-11-13
Be_between matcher

Pazus:
Be_between matcher implementation.
Tests are included.

On review please pay attention to the null processing, not sure I've implemented it correctly as designed. What if expected value is null, what if actual is null.

### DIFF
--- a/source/expectations/matchers/be_between.tpb
+++ b/source/expectations/matchers/be_between.tpb
@@ -1,0 +1,92 @@
+create or replace type body be_between is
+
+  member procedure init(self in out nocopy be_between, a_lower_bound ut_data_value, a_upper_bound ut_data_value) is
+  begin
+    self.name            := lower($$plsql_unit);
+    self.lower_bound     := a_lower_bound;
+    self.upper_bound     := a_upper_bound;
+    self.additional_info := 'between ' || a_lower_bound.to_string || ' and ' || a_upper_bound.to_string;
+  end;
+
+  constructor function be_between(self in out nocopy be_between, a_lower_bound date, a_upper_bound date)
+    return self as result is
+  begin
+    init(ut_data_value_date(a_lower_bound), ut_data_value_date(a_upper_bound));
+    return;
+  end;
+  constructor function be_between(self in out nocopy be_between, a_lower_bound number, a_upper_bound number)
+    return self as result is
+  begin
+    init(ut_data_value_number(a_lower_bound), ut_data_value_number(a_upper_bound));
+    return;
+  end;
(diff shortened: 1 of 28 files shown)


Pazus:
I'll come back and fix the issue

jgebal:
Can you please add support for varchar2 as well (for completeness)?
This code is perfectly valid in PLSQL so maybe asserts should also support it:
```sql
begin
  if 'b' between 'a' and 'cb' then
    dbms_output.put_line('ok');
  end if;
end;
/
```

tests/asssertions/ut.expect.to_be_between.GivesSuccessWhenDifferentTypes.sql
```diff
@@ -0,0 +1,5 @@
+PROMPT Gives failure when expected are numbers and and actual is varchar
```
jgebal: the prompt is not correct.

tests/asssertions/ut.expect.to_be_between.GivesSuccessWhenDifferentTypes.sql
```diff
@@ -0,0 +1,5 @@
...
+@@asssertions/common/ut.expect.to_be_between.scalar.different_types.common.sql 'varchar2(4000)' 'number' '''1''' '0' '2'
+
+PROMPT Gives failure when expected are varchars and and actual is number
```
jgebal: and here

source/expectations/matchers/be_between.tpb
```diff
@@ -58,6 +64,14 @@ create or replace type body be_between is
       begin
         l_result := l_actual.datavalue between l_lower.datavalue and l_upper.datavalue;
       end;
+    elsif self.lower_bound is of(ut_data_value_varchar2) and self.lower_bound is of(ut_data_value_varchar2) and a_actual is of(ut_data_value_varchar2) then
```
jgebal: typo lower_bound is checked twice and upper_bound is not checked.

source/expectations/ut_expectation_varchar2.tps
```diff
@@ -1,6 +1,7 @@
 create or replace type ut_expectation_varchar2 under ut_expectation
 (
   overriding member procedure to_equal(self in ut_expectation_varchar2, a_expected varchar2, a_nulls_are_equal boolean := null),
+  member procedure to_be_between(self in ut_expectation_varchar2, a_lower_bound varchar2, a_higher_bound varchar2),
```
jgebal: maybe we can use `upper_bound` not `higher_bound`

Pazus: done

source/expectations/matchers/be_between.tpb
Pazus: fixed!

Issue #85, opened 2016-10-18
Refactor design of reporters to separate format from output stream


jgebal:
Streaming to output DBMS OUTPUT/PIPE/FILE/AQ should be separated from the reporter.
Reporter should only deal with formatting.
The `ut_output` sub classes should be responsible for placing outcomes somewhere.
This will greatly improve flexibility of reporting.

PR #104, opened 2016-11-13
Feature/separate output from reporter

jgebal:
Output is now separated from reporter.
A new interface `ut_output` is created.
A new `ut_output_dbms_output` class is created.
Method print is moved from into the base class `ut_reporter`.
Refactored `ut_teamcity_reporter_printer` to become `ut_teamcity_reporter_helper`.
Now it is a set of functions returning the outcomes to be printed.
This allows the reporter to keep control over the output used.


jgebal:
@pazus.
Can you review the changes? I'm interested in your opinion around teamcity package function change (to be used as helper)

source/core/types/ut_output_dbms_output.tpb
```diff
@@ -0,0 +1,34 @@
...
+    null;
+  end;
+
+  overriding member procedure send(self in out nocopy ut_output_dbms_output, a_text clob) is
+    l_text_part varchar2(32767 byte);
+    --we're assuming max of 2 bytes per char
+    c_size_limit_chars constant integer := (32767/2);
+    i integer := 0;
+  begin
+    while i <= length(a_text) loop
```
Pazus: I think here "less" in neede instead of "less or equal" as if the string length is equal to `c_size_limit_chars` an additional empty line will appear.

source/core/ut_suite_manager.pkb
```diff
@@ -290,15 +292,16 @@ create or replace package body ut_suite_manager is
       l_ind := g_schema_suites(a_owner_name).first;
       while l_ind is not null loop
         l_suite := g_schema_suites(a_owner_name) (l_ind);
-        l_suite.do_execute(a_reporter => a_reporter);
+        a_reporter := l_suite.do_execute(a_reporter => a_reporter);
```
Pazus: Decided to leave the only implementation of `do_execute` as procedure with `in no nocopy` reporter parameter.

Pazus:
I like your rename or `ut_teamcity_reporter_printer`, "helper" makes more sence with configurable output approach.

source/core/types/ut_output_dbms_output.tpb
jgebal: good pint - will fix it now

source/core/ut_suite_manager.pkb
jgebal: As per discussion i'll remove the overloading (procedure/function) and we will use procedure (in out nocopy)

jgebal:
All comments adressed.
This PR resolves #85

PR #108, opened 2016-11-19
parse suitepackages bug fix.

GolezTrol:
ltrim just trims all given characters, it doesn't check patterns.
example: when specifying %suitepackages(all.logisics.xyz), the l from logistics would be trimmed as well.


jgebal:
Hi @goleztrol, can too you add unit tests th hat would document how the change works?
Best would be a test or set of tests that failed before change and now pass.
@pazus, it would be great to get your comments on this change as you are most familiar with the parser.

Issue #103, opened 2016-11-13
Line and package of the failed assertion


Pazus:
Store the line and the package name at which the assertion failed. So it can be used by reporter for detailed info about the failure.

Issue #101, opened 2016-11-12
Change ut_dbms_output_suite_reporter to ut_documentation_reporter


jgebal:
The reporter should provide text outputs in a format as below image:
![image](https://cloud.githubusercontent.com/assets/1398171/20240825/1fef1db4-a91a-11e6-8d21-42ada82d2396.png)

The items of high importance
1. Suites are reported by descriptive name (if avaliable)
Each sub-item (suite or test) is indented by with 2 spaces
2. There is a single empty line present between individual suites
3. If a test is failing, a text `(FAILED - x)` is shown next to the test name. `x` is a consecutive failure number
4. If suite is not named then a package name with owner name is used
If a test is not named, then a test procedure name is used (with brackets)
5. After all the tests were reported, and if any of tests failed, a `Failures:` section is present.
The section is at the bottom of the report, appearing only once, one for all suites executed in a run.
6. The `(FAILED - x)` tests are referenced in `Failures:` report in order of failure.
For each failed test a **full name** is built out of: root suite name, branch suites, test name.
For each failed test, a `Failures/errors:` section is shown.
The section contains information about all failed assertions.
The information contains information that will allow identifying why the assertion and test failed.
7.  For each failed assertion, a `owner.package:line_number` is used to help navigate to the failure
8. At the end of the report, information about run time 
9. Information about number of tests. IF any failed, number of failed is displayed too.
10. Brief summary of failed examples in a format:
owner.package_name:line_number # **full name**

jgebal:
This is resolved by #109 
The formatting of  reporter output is documented by examples
```
Demo of documentation reporter
  repoting exception (FAILED - 1)
  UT3.DEMO_DOC_REPORTER1.failing_no_name (FAILED - 2)
  A failing test exsample (FAILED - 3)
  A passing test sample
  UT3.DEMO_DOC_REPORTER1.test_without_name

A suite pacakge without body
  A test (FAILED - 4)
  UT3.DEMO_DOC_REPORTER2.passing_test2 (FAILED - 5)

suite_package_without_name
  A passing test sample
  A passing test sample

Failures:
   1) repoting exception
      Failures/Errors:
	error: ORA-01858: a non-numeric character was found where a numeric was expected
	       ORA-06512: at "UT3.DEMO_DOC_REPORTER1", line 21
	       ORA-06512: at line 5

   2) UT3.DEMO_DOC_REPORTER1.failing_no_name
      Failures/Errors:
[...]

PR #109, opened 2016-11-25
Added documentation-like reporter.

jgebal:
Minor refactorings done.
Fixed suite manager to call `after_run` with a list of executed suites.


source/reporters/ut_documentation_reporter.tpb
```diff
@@ -0,0 +1,130 @@
...
+    l_suite ut_test_suite := treat(a_suite as ut_test_suite);
+  begin
+    lvl := lvl - 1;
+    self.print( ' ' );
+  end;
+
+  overriding member procedure after_run(self in out nocopy ut_documentation_reporter, a_suites in ut_objects_list) as
+    l_start_time    timestamp with time zone := to_date('9999','yyyy');
+    l_end_time      timestamp with time zone := to_date('0001','yyyy');
+    procedure print_failures_summary is
```
Pazus: I believe this should be moved out from the reporter. Maybe we have to define a "run_summory" type that contains such information so the reporter can just present it.

source/core/types/ut_executable.tpb
```diff
@@ -1,10 +1,10 @@
 create or replace type body ut_executable is
 
-  static procedure execute_call(a_owner varchar2, a_object varchar2, a_procedure_name varchar2) is
-    l_stmt varchar2(200);
-    l_cursorid    number;
-    l_rowsprocessed    number;
-  
+  static function execute_call(a_owner varchar2, a_object varchar2, a_procedure_name varchar2) return varchar2 is
```
jgebal: Good idea. This will make the design a bit better.

source/core/types/ut_executable.tpb
```diff
@@ -13,20 +13,34 @@ create or replace type body ut_executable is
...
+    'begin' || chr(10) ||
+    '  begin' || chr(10) ||
+    '    ' || ut_metadata.form_name(l_owner, l_object_name, l_procedure_name) || ';' || chr(10) ||
+    '  exception' || chr(10) ||
+    '    when others then ' || chr(10) ||
+    '      --raise on ORA-04068: existing state of packages has been discarded to avoid unrecoverable session exception' || chr(10) ||
+    '      if sqlcode = -04068 then' || chr(10) ||
+    '        raise;' || chr(10) ||
+    '      end if;' || chr(10) ||
+    '      l_error_stack := dbms_utility.format_error_stack || dbms_utility.format_error_backtrace;' || chr(10) ||
```
jgebal: yup

source/reporters/ut_documentation_reporter.tpb
jgebal: I agree, that is not a best place for this code, as it can be reused.
On the other hand, we can always create reporters on top of this reporter.
Maybe a package `ut_documentation_reporter_helper`?
I can't think of a right place to put this code in.
Anyway, for now I've refactored the procedure and split it into several pieces to separate smaller functionalities.

jgebal: If you could modify the code according to what you think would be appropriate, please do so.

Pazus:
@jgebal I have looked through the changes you made from my review.

I'd like to offer next moves:
1) reorgonize the branch a bit to make the PR a smaller and more focused on the new reporter. Move all other changes to another branch as I think we need to work them through before merging.
2) think I'd like to be worked are:
a) Examples. we discussed earlier that we need to reorganize examples section to be more for user then for the framework developer. From that point some changes you've made are not needed, but others are lacked. 
b) I don't see a reason to remove the `ut_custom_reporter` example. It shows the way a user can define he's own reporter. Maybe we should update the approach and the example with the use of outputters etc.
c) I remember we talked about it and I think all examples with manual suite construction should be gone after we implement the `ut.run`, so I think that changes should be postponed a bit.
d) and on the В item i realized you've moved examples to the `examples_for_extension_developers` :) That's the absolutely right change but I think it should be moved to another branch.
e) So i'd leave only changes made to examples the correspond deletion of dbms_reporter and extract all the others to a separate PR.

The changes to the main PR files are fine.

jgebal:
Hi @Pazus 
As per your suggestion I have removed the changes to the examples.

jgebal:
This PR resolves #103

jgebal:
Resolves #101 
Resolves #110

jgebal:
@Pazus do you think i should also separate the changes to executors and put them into a separate PR?
Not sure if that would be of any value except for separation of commits and merges.

source/core/types/ut_assert_result.tpb
```diff
@@ -7,6 +7,7 @@ create or replace type body ut_assert_result is
     self.object_type   := 0;
     self.result        := a_result;
     self.error_message := a_error_message;
+    self.caller_info   := ut_assert_processor.who_called_expectation();
```
jgebal: for performance reasons, we could limit those calls to only occur when assert failed.
Also, maybe we should rename the type to be `ut_expectation_result`?

Pazus: agree!

Issue #74, opened 2016-10-17
Running Unit Tests from different schema.


jgebal:
As part of this feature we need to:
- include installation option to:
  - create synonyms (public/specific user) for public API elements
  - add grants (public/specific user) to public API objects
- within Travis CI installation test the public/specific user option by:
  - creating the 'specific user' so that the user uses private synonyms 
  - creating 'another user' so that the user uses public synonyms
  - installing a unit test package in 'specific user' schema and running it
  - installing a unit test package in 'another user' schema and running it

PR #112, opened 2016-11-29
non ut3 user

Shoelace:
this PR have teh follwoing changes
* creates a secondary user ,assigns grants, create synoyms and reruns tests via non-owner user
* modified get_package_spec_source to use dbms_preprocessor instead


source/core/ut_metadata.pkb
```diff
@@ -134,12 +104,28 @@ create or replace package body ut_metadata as
...
   begin
-    l_cur := get_package_spec_source_cursor(a_owner, a_object_name);
-    fetch l_cur bulk collect into l_txt_tab;
-    close l_cur;
-    return ut_utils.table_to_clob(l_txt_tab);
-  end;
+    dbms_lob.createtemporary(l_source, true);
+    
+
+      l_source_lines := SYS.DBMS_PREPROCESSOR.GET_POST_PROCESSED_SOURCE(
```
Pazus: we defined lowecase in the convention

Shoelace: true.. i just cut and paste that.  so thats easy to fix..

jgebal:
Relates to #74

examples/tst_pkg_huge.pks
```diff
@@ -305,6 +305,8 @@ create or replace package tst_pkg_huge is
     -- %test(Test number 100)
     procedure test100;
     
+$IF $$tst_slim $THEN
```
Shoelace: well the point was to show/test that the get_source only get the correct CC compiled code..  so if you have packages under test that use conditional compilation it gets tested correctly still.. but sure.. it should probably be moved to test.. and not example.

source/core/ut_metadata.pkb
```diff
@@ -134,12 +104,24 @@ create or replace package body ut_metadata as
   function get_package_spec_source(a_owner varchar2, a_object_name varchar2) return clob is
     l_txt_tab ut_varchar2_list;
     l_cur     sys_refcursor;
+    l_source  clob;
```
Shoelace: yep fine.. good point. i'll re-add the back the blank lines.  the `get_package_spec_source` function is of course only returning the package spec. but the preprocessor can return the body.  as to which line numbers are returned we'd have to test.. but i would assume lines refer to the unprecessed source.. but since teh preprocessor just replaces with blanks that shoudl be fine still

source/core/ut_output_pipe_helper.pks
```diff
@@ -1,5 +1,7 @@
 create or replace package ut_output_pipe_helper is
 
+  gc_enabled constant boolean := $if $$ut_pipe_enable $then TRUE; $else FALSE; $end
```
Shoelace: we it was going to eb needed to know if dbms_pipe was enabled or not.. but since we're takign out teh $cc then this can go too.. yes.

source/core/types/ut_output_dbms_pipe.tpb
```diff
@@ -13,7 +13,11 @@ create or replace type body ut_output_dbms_pipe as
     --create an explicit private pipe. Explicitly created pipes need to be removed explicitly
     --otherwise they stay in memmory forever https://docs.oracle.com/cd/B19306_01/appdev.102/b14258/d_pipe.htm#CHDEICJI
     --to check if there are any non-purged pipes execute:  select * from v$db_pipes where pipe_size > 0 order by name desc;
+$if $$ut_pipe_enable $then
```
jgebal: Can you change the install, so that the pipe output is installed optionally for now?
Either by separate script (that still executes on Travis) or by parametrizing the install script.

source/core/ut_metadata.pkb
jgebal: OK, so we should be all good.

jgebal:
The more I think about it the more i like the idea of using preprocessor.
By using it for fetching sources to parse, we avoid situation that UT contain conditional compilation and our engine is not aware of it and tries to execute tests that are actually disabled by compilation.

It is definitely worth a simple unit test.

source/core/ut_metadata.pkb
jgebal: I get your point.
We probably should not be using preprocessor for lookup of package body source line, that was called, though we could, but would need to fetch whole source just to get one line - not very optimal.

examples/tst_pkg_huge.pks
jgebal: Definitely worth a simple unit test: "Only execute the code that was not disabled by conditional compilation"

examples/RunAllExamples.sql
```diff
@@ -18,5 +21,7 @@ prompt RunExampleTestAnnotationBasedForCurrentSchema
 @@RunExampleTestAnnotationBasedForCurrentSchema.sql
 prompt RunExampleTestAnnotationsHugePackage
 @@RunExampleTestAnnotationsHugePackage.sql
+prompt RunExampleTestAnnotationsNotsoHugePackage
+@@RunExampleTestAnnotationsNotsoHugePackage.sql
```
jgebal: Now I get it. I erase my case, though it should be a UT not an example.

.travis/utplsql_synonyms.sql
```diff
@@ -0,0 +1,48 @@
...
+set echo off
+set feedback on
+set heading off
+set verify off
+
+define ut3_owner   = "&1"
+define ut3_user    = "&2"
+define ut3_syntype = "&3"
+
+create &ut3_syntype synonym &ut3_user.ut_assert                       for &ut3_owner..ut_assert;
```
jgebal: Nice.
That was actually an ask form one of DBA's, to make sure that SYNONYMS do not use CREATE OR REPLACE so that we do not have a risk of overwriting synonyms for existing objects if there are conflicts of names.

.travis/create_utplsql_user.sh
```diff
@@ -4,5 +4,15 @@ set -ev
...
+SQL
+#give grants
+"$ORACLE_HOME/bin/sqlplus" -L -S $UT3_OWNER/$UT3_OWNER_PASSWORD <<SQL
+set echo off
+@@$(dirname "$(readlink -f "$0")")/utplsql_grants.sql $UT3_USER 
+SQL
+# create synonyms
+"$ORACLE_HOME/bin/sqlplus" -L -S / as sysdba <<SQL
+set echo off
+@@$(dirname "$(readlink -f "$0")")/utplsql_synonyms.sql $UT3_OWNER "$UT3_USER." ""
```
jgebal: Can you make sure that we can simply pass `$UT3_USER` without "dot" as a parameter?

.travis/utplsql_synonyms.sql
```diff
@@ -0,0 +1,48 @@
...
+create &ut3_syntype synonym &ut3_user.ut_test                       for &ut3_owner..ut_test;
+create &ut3_syntype synonym &ut3_user.ut_reporter                   for &ut3_owner..ut_reporter;
```
jgebal: Ah, now i know why you need dot in the parameter value.
That is a bit hacky though it works.

examples/demo_expectations.pck
```diff
@@ -23,10 +23,10 @@ create or replace package demo_expectations is
   -- %test(demo of failure for to_be_null expectation )
   procedure demo_to_be_null_failure;
 
-  -- %test(demo of failure for to_be_null expectation )
+  -- %test(demo of success for to_be_null expectation )
```
jgebal: thanks!

Shoelace:
i think i have resolved all of the issues from your review @jgebal but there is a dbms_pipe error in the travis run that is failing. i think it should be sorted before being merged.

Pazus:
I close this PR as outdate, work will be continued in #152

Issue #86, opened 2016-10-18
Create single entry procedure/function for running tests


jgebal:
Currently suite_manager has several ling-named procedures to run ut's.
Can we have just one name as an entry point?

Pazus:
Do you mean a similary named procedures dirrefer by parameter types?

jgebal:
Hi @pazus 
What I was thinking of is to define a consistent, simple and flexible way for developers to execute tests.
Looking at most ut frameworks, they offer single-line or single command ut invocation.

I would like to see something similar in utplsql.
The inputs for a runner should be probably:
- what to run: suite/suites/ut package/ a test under carret in package (for IDE integrations)/a test procedure in package
- what to skip (optionally)
- what should be the report formats (reporters)
- what should be the report outputs
- default transaction control for run
- default null comparison for run
- more??

What we define initially should be created in a way that will be easy to extend.

Pazus:
@Jgebal, now I see your point!

So we should have procedures `run_singlr_suite`, `run_schema_suites` (run test is arguable due to suite setup may be required).

Main parameters. Run_single_suite: `ut_test_suit`e or owner/package name or `suite_path`.
Run_schema_suites: schema name.

Both procedures accept reporter(which contains out-putter) to use and list of tags to include(all of empty)/exclude (none if empty).

I don't think transaction control and null comparison should be configurable on execution. Imagine I have a suite package with several tests and suite setup. If it is designed with rollback after test in mind then all the test assume that the data at its start is the same as after the setup procedure. If I run such suite with overridden transaction control most likely the second and the following tests will behave  
wrong.
The same is true for null comparison. I think this behavior is set on design.

Pazus:
May we leave only the singe entry point passing for this issue as it is crucial to have it before Alfa and postpone passing multiple entry points to another Issue with another milestone?

Pazus:
Single entry point is implemented for a single-path argument, merged by PR #113

Pazus:
Moved the milestone to 3.1. I think multipath implementation won't be done before that.

jgebal:
I think we can consider this task close, despite the fact the multipath runner is not complete. API is there which was the expectation for this task.

PR #113, opened 2016-11-29
Implemented the runner to execute suites/tests by path

Pazus:
Implemented the runner to execute suites/tests by path

The path supports several formats:
* "schema"
* "schema":"suitepath"[."test_proc_name"]
* "schema"."suite_package"[."test_proc_name"]

The user-freandly interface is located in `ut` package

Also made some refactoring with ut_test objects, put object_path property there to store fully qualified path to suite/test
object name now contains only the name of the object (package/procedure)


source/core/ut_suite_manager.pkb
```diff
@@ -327,5 +339,166 @@ create or replace package body ut_suite_manager is
...
+        if regexp_like(l_path, '^\w+(\.\w+){1,2}$') then
+          declare
+            l_temp_suite     ut_test_suite;
+            l_package_name   varchar2(4000);
+            l_procedure_name varchar2(4000);
+          begin
+            l_package_name   := regexp_substr(l_path, '^\w+\.(\w+)(\.(\w+))?$', subexpression => 1);
+            l_procedure_name := regexp_substr(l_path, '^\w+\.(\w+)(\.(\w+))?$', subexpression => 3);
+          
+            l_temp_suite := config_package(l_schema, l_package_name);
```
jgebal: Do we need to fetch/parse the package again? We already have whole schema suites in variable in line 438.

source/core/ut_suite_manager.pkb
```diff
@@ -327,5 +339,166 @@ create or replace package body ut_suite_manager is
...
+            l_path       := rtrim(l_schema || ':' || l_temp_suite.object_path || '.' || l_procedure_name, '.');
+          end;
+        end if;
+        
+        if regexp_like(l_path, '^\w+:.+$') then
```
jgebal: should this be elsif?

source/core/ut_suite_manager.pkb
```diff
@@ -327,5 +339,166 @@ create or replace package body ut_suite_manager is
...
+          raise_application_error(-20203, 'Suite note found');
+        end if;
+      end if;
+    end set_skipped_flag;
+  begin
+    l_paths := clean_paths(a_paths);
+  
+    validate_paths(l_paths);
+  
+    -- current implementation operates only on a single path
```
jgebal: Maybe it would be better to extract the find/filter logic into separate procedure?

source/core/ut_suite_manager.pkb
```diff
@@ -327,5 +339,166 @@ create or replace package body ut_suite_manager is
...
+          l_objects_to_run.extend;
+          l_objects_to_run(l_objects_to_run.count) := l_suite;
+        end if;
+
+      end if;
+    
+    end loop;
+  
+    if l_objects_to_run.count > 0 then
+      l_reporter.before_run(a_suites => l_objects_to_run);
```
jgebal: Nice.
Now we pass the full list of objects to execute to the before/after calls.

jgebal:
Resolves #86

source/core/ut_suite_manager.pkb
```diff
@@ -118,46 +122,47 @@ create or replace package body ut_suite_manager is
...
-                                   --  ut_utils.gc_rollback_on_error
-                                   else
-                                     l_suite_rollback
-                                   end;
+              l_rollback_type       := case lower(l_rollback_annotation)
+                                         when 'manual' then
+                                          ut_utils.gc_rollback_manual
+                                         when 'auto' then
+                                          ut_utils.gc_rollback_auto
+                                       --when 'on-error' then
```
Pazus: not now. this remindes about the third type of transaction control. it might became possible to implement after your changes to `ut_executable`

source/core/ut_suite_manager.pkb
```diff
@@ -327,5 +339,166 @@ create or replace package body ut_suite_manager is
...
+  
+    return g_schema_suites(a_schema_name);
+  end get_schema_suites;
+
+  -- Validate all paths are correctly formatted
+  procedure validate_paths(a_paths in ut_varchar2_list) is
+    l_path varchar2(32767);
+  begin
+    if a_paths is null or a_paths.count = 0 then
+      raise_application_error(ut_utils.gc_path_list_is_empty, 'Path list is empty');
```
Pazus: though about it but for now decided not to do it as it get much more difficult to figure out what is really called. the path will always be in form `suite1.suite2.test` and we have to understant wether its a suitepath of current schema ot suite1 is the schema name. Guessing mechanism is not yet implemented

source/core/ut_suite_manager.pkb
```diff
@@ -327,5 +339,166 @@ create or replace package body ut_suite_manager is
...
+    else
+      for i in 1 .. a_paths.count loop
+        l_path := a_paths(i);
+        if regexp_like(l_path, '^\w+(\.\w+){0,2}$') or regexp_like(l_path, '^\w+:\w+(\.\w+)*$') then
```
Pazus: ok.

source/core/ut_suite_manager.pkb
Pazus: i feel such separation won't work for multiple pathes provided as be each next path you won't simply add new object to the list, rather you will enable objects in current list to execute... I'd like to postepone this refactoring a bit until it became clear for multi-path approach. But I agree it is needed

source/core/ut_suite_manager.pkb
Pazus: Then we will have to work through the whole hierarcy looking for `object_name`... but maybe you are right. i'll come back to it today a bit later.

jgebal:
@Pazus do you think i should also separate the changes to executors and put them into a separate PR?
Not sure if that would be of any value except for separation of commits and merges.

Pazus:
Added tests

Issue #114, opened 2016-12-02
Smart re-parsing of existing packages


GolezTrol:
`ut_suite_manager.run_cur_schema_suites_static` has this nice little parameter called `a_force_parse_again`. However, it seems to be all or nothing.

When I started to use utPLSQL, I found out that I created one test package for a unit of code (another package) that I wanted to test, and kept adding tests to that test package.

I think utPLSQL could be 'smart' in detecting whether or not I recompiled the package header of my test package since the last time it was parsed using `USER_OBJECTS.LAST_DDL_TIME`, and only in that case re-parse the package header.
I would still have to manually force reparsing when I add a new test suite, but changes to a suite or even dropping one could be detected this way.

This would cover the most common scenario of having a unit to test, having a test suite for it, and just making changes and additions to the test.

Benefits:
- Speed up running the tests, by only reparsing when needed.
- Speed up development, by automatically reparsing when needed.

jgebal:
That was done in annotations cache and probably could be added to the
project.
I think what we could do is create an interface for parser so we can decide
if we want to install and use cache or not.
@oasis. What do you think?

oasis:
What do I think? I think you've tagged the wrong guy! =P

jgebal:
@oasis Haha. Auto-correct on Android is playing tricks with me.
It was meant for @pazus

jgebal:
@GolezTrol Can you have a look at #57 - it was closed and not merged but I'm thinking about giving it a second chance with a bit different approach on integrating it.

Pazus:
Actually we do have cache right now. When the schema is parsed all it's suites are stored in `ut_suite_manager` cache per schema. The problem is 1) it doesn't invalidates on DDL, 2) it is not smart, the whole schema is parsed.

GolezTrol:
True. So instead of having `a_force_parse_again=Y/N`, I would be looking for `a_reparse=never/always/cache_changed/package_changed` or so. I'd imagine these possibilities:

1.  None: Just don't update the cache once the cache is made.
1.  Always/Force: Always update the entire cache, regardless.
1.  Cached package change: Check the last DDL of the packages that were cached before. Only reparse those packages when they have changed. Remove deleted test packages from the list.
1.  Any package change: Check the last DDL of all packages, like above, reparse modified test packages, but also, if there is a package that was added or modified, but wasn't cached before, parse it to see if it is a test package.

[...]

PR #117, opened 2016-12-07
Smart schema re-parse

Pazus:
Issue #114 in a simplified form.
Now schema cache becomes invalidated when the last_ddl_time of any package in the schema change
Also corrected examples to eliminate calls to deprecated procedures:
`run_schema_suites`
`run_schema_suites_static`
`run_cur_schema_suites`
`run_cur_schema_suites_static`

### DIFF
--- a/source/core/ut_suite_manager.pkb
+++ b/source/core/ut_suite_manager.pkb
@@ -1,10 +1,10 @@
 create or replace package body ut_suite_manager is
 
-  type tt_schema_suits is table of ut_test_suite index by varchar2(4000 char);
-  type t_schema_cache is record (schema_suites tt_schema_suits, changed_at date);
-  type tt_schena_suits_list is table of t_schema_cache index by varchar2(32 char);
+  type tt_schema_suites is table of ut_test_suite index by varchar2(4000 char);
+  type t_schema_cache is record (schema_suites tt_schema_suites, changed_at date);
+  type tt_schena_suites_list is table of t_schema_cache index by varchar2(32 char);
 
-  g_schema_suites tt_schena_suits_list;
+  g_schema_suites tt_schena_suites_list;
 
   function trim_path(a_path varchar2, a_part varchar2) return varchar2 is
   begin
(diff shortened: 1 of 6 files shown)


source/core/ut_suite_manager.pkb
```diff
@@ -290,62 +306,16 @@ create or replace package body ut_suite_manager is
...
-
   function get_schema_suites(a_schema_name in varchar2) return tt_schema_suits is
   begin
     -- Currently cache invalidation on DDL is not implemented so schema is rescaned each time
-    --if not g_schema_suites.exists(a_schema_name) then
-    --  config_schema(a_schema_name);
-    --end if;
-    config_schema(a_schema_name);
+    if not g_schema_suites.exists(a_schema_name) 
+      or g_schema_suites(a_schema_name).changed_at<nvl(get_schema_max_ddl(a_schema_name), date'2999-12-31') then
```
jgebal: I would put the NVL into the `get_schema_max_ddl`.

source/core/ut_suite_manager.pkb
```diff
@@ -1,14 +1,26 @@
 create or replace package body ut_suite_manager is
 
   type tt_schema_suits is table of ut_test_suite index by varchar2(4000 char);
```
jgebal: typo in type name here and in line 5

source/core/ut_suite_manager.pks
```diff
@@ -3,14 +3,6 @@ create or replace package ut_suite_manager authid definer is
...
-  procedure run_schema_suites(a_owner_name varchar2, a_reporter in out nocopy ut_reporter, a_force_parse_again boolean default false);
-
-  procedure run_schema_suites_static(a_owner_name varchar2, a_reporter in ut_reporter, a_force_parse_again boolean default false);
-
-  procedure run_cur_schema_suites(a_reporter in out nocopy ut_reporter, a_force_parse_again boolean default false);
-
-  procedure run_cur_schema_suites_static(a_reporter in ut_reporter, a_force_parse_again boolean default false);
   
   procedure run(a_path in varchar2, a_reporter in ut_reporter);
   -- to be implemented
```
jgebal: I would enable this run procedure anyway (even if it only runs for one element of array at the moment).
This way we have a complete API and can start implementing on top of it.
Internal implementation can change for `run( list )` later. But if we're OK with the entry point signature, it should be enabled I think.

PR #123, opened 2016-12-14
Feature/better examples

jgebal:
- Added award_bonus example
- Added betwnstr example
- Added remove rooms example

Those are meant to be real use-case scenarios (a bit vanilla but still usable)

### DIFF
--- a/examples/RunAllExamples.sql
+++ b/examples/RunAllExamples.sql
@@ -21,9 +21,12 @@ prompt RunExampleTestAnnotationsHugePackage
 @@RunExampleTestAnnotationsHugePackage.sql
 prompt RunExpectations
 @@RunExpectations.sql
-prompt RunWithDocumentationReporter
+
 @@RunWithDocumentationReporter.sql
 
+--workaround for suite level cache
+declare d date := sysdate; begin loop exit when d != sysdate; end loop; end;
+/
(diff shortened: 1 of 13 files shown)


examples/RunAllExamples.sql
```diff
@@ -21,5 +21,18 @@ prompt RunExampleTestAnnotationsHugePackage
 @@RunExampleTestAnnotationsHugePackage.sql
 prompt RunExpectations
 @@RunExpectations.sql
-prompt RunWithDocumentationReporter
+
 @@RunWithDocumentationReporter.sql
+
+--workaround for suite level cache
+declare d date := sysdate; begin loop exit when d != sysdate; end loop; end;
```
Pazus: cache invalidation is fixed in another PR, no more need in that

examples/remove_rooms_by_name/rooms.sql
```diff
@@ -0,0 +1,12 @@
+create table rooms (
+  room_key number primary key,
+  name varchar2(100)
+);
+
+create table room_contents (
+  contents_key number primary key,
+  room_key number,
```
Pazus: This is a bit simpler:
```
room_key number references rooms(room_key),
```

jgebal:
I've addressed the review comments.

PR #129, opened 2016-12-18
Feature/running with many reporters

jgebal:


jgebal:
Hi @pazus. I've udated the PR with changes and addressed your comment.

jgebal:
I'm just thinking if the ut_run_sqlplus.sql should be in `/sources` or do we need a new directory called `/client` to store examples of client side code used for invoking UTPLSQL with all it's power and reporting capabilities.

Pazus:
How about placing it in the root?

source/core/types/ut_output_stream.tpb
```diff
@@ -23,3 +23,8 @@ create or replace type body ut_output_stream as
 
 end;
 /
+--we need to use plsql optimize level=1 to prevent
```
Pazus: How about changing the " for loop" to open-loop-fetch-close? I thought optimization leaves it as is. This way altering plsql_optimizate_level is not needed (but should be rechecked)

jgebal: @pazus Can you do that change?

PR #136, opened 2016-12-24
Sorted tests in suites in the order they appear in package

Pazus:
Shuffled execution will be implemented later


jgebal:
Looks good, thanks @Pazus.

Issue #137, opened 2016-12-26
Fail all tests if setup of suite fails


Pazus:
If a failure occurs in the setup of a suite, tests are still executed. That is wrong.

jgebal:
I don't think it's wrong.
Tests that actually depend on the setup will naturally fail. Tests that ate independent will succeed. That is fine.
We would however intercept the exceptions from. Before/after block and give a warning

jgebal:
Do you think that printing the exceptions that were raised in before/after block would be enough?

Pazus:
I've faced that issue. Here is what you see. You have a set of tests, they all passes". Then after some change you have half of them failing with strange errors. It takes time to figure out that actually hidden error in the setup occurred.

Though I like your idea that the warning approach makes sense. I think we need a draft.

jgebal:
Documentation reporter should use after_setup and after_teardown hooks to report errors.

Pazus:
Then we need a way to store info that error occurred. There is no such place now.

Pazus:
@jgebal Even after refactoring the problem persists. If `beforeall` fails, then no items are executed and calc_execution_result procedure marks suite as successful.
We need to find a way to fix that.

PR #138, opened 2016-12-26
Implemented suite failure if setup procedure fails

Pazus:
Implemented suite failure if setup procedure fails
Looking for idea what to do when suite's teardown procedure fails


Pazus:
addresses #137

Pazus:
Don't merge for now. I implemented that before I read @jgebal comment.

jgebal:
I think we should drop this PR.

Pazus:
Need another solution and rewrite

Issue #133, opened 2016-12-23
Reorganize annotations


Pazus:
Lets reogranize all our annotations to match common approaches implemented in jUnit 5 and RSpec. This will close the framework to other testing fremawork out there and make learning easier for other programmers.

Here are guiding documents we should keep in minds:
NUnit 3: https://github.com/nunit/docs/wiki/Attributes
JUnit 5: http://junit.org/junit5/docs/current/user-guide/#writing-tests-annotations
RSpec : (@jgebal  please provide correct link)

| Annotation |Level| Describtion |
| --- | --- | --- |
| `%suite` | Package | Marks package to be a suite of tests This way all testing packages might be found in a schema. |
| `%suitepath(<path>)` | Package | Similar to java package. The annotation allows logical grouping of suites into hierarcies. |
| `%displayname(<description>)` | Package/procedure | Human-familiar describtion of the suite/test. `%displayname(Name of the suite/test)` |
| `%test` | Procedure | Denotes that a method is a test method. |
| `%beforeall` | Procedure | Denotes that the annotated procedure should be executed before all `%test` methods in the current suite. |
| `%afterall` | Procedure | Denotes that the annotated procedure should be executed after all `%test` methods in the current suite. |
| `%beforeeach` | Procedure | Denotes that the annotated procedure should be executed before each `%test` method in the current suite. |
| `%aftereach` | Procedure | Denotes that the annotated procedure should be executed after each `%test` method in the current suite. |
| `%beforetest(<procedure_name>)` | Procedure | Denotes that mentioned procedure should be executed before the annotated `%test` procedure. |
| `%aftertest(<procedure_name>)` | Procedure | Denotes that mentioned procedure should be executed after the annotated `%test` procedure. |
| `%rollback(<type>)` | Package/procedure | Configure transaction control behaviour (type). Supported values: `auto`(default) - rollback to savepoint (before the test/suite setup) is issued after each test/suite teardown; `manual` - rollback is never issued automatically. Property can be overridden for child element (test in suite) |
| `%disable` | Package/procedure | Used to disable a suite or a test |
| `%tag(<tag>)` | Package/procedure | Used to declare _tags_ for filtering tests, either at the suite or test level. |
[...]

PR #145, opened 2017-01-08
Annotations redesign

Pazus:
Redesign according to #133 

Changin code to use new `%displayname` annotation a realized I like the old way `%suite(Name of the suite)`, `%test(Name of the test)` really much, so maybe we should get back to the old approach.
Additionaly as IDE doesn't help us with autocomplete on annotations, the less of them we need to specify, the more stable is test package configuration.
What do you think? Look at the comparison of old and new examples.

Or we can use both aproaches with `%displayname` prioritized.

### DIFF
--- a/docs/userguide/annotations.md
+++ b/docs/userguide/annotations.md
@@ -56,8 +56,8 @@ end test_pkg;
 | `%suitepath(<path>)` | Package | Similar to java package. The annotation allows logical grouping of suites into hierarcies. |
 | `%displayname(<description>)` | Package/procedure | Human-familiar describtion of the suite/test. `%displayname(Name of the suite/test)` |
 | `%test` | Procedure | Denotes that a method is a test method. |
-| `%beforeall` | Procedure | Denotes that the annotated procedure should be executed before all `%test` methods in the current suite. |
-| `%afterall` | Procedure | Denotes that the annotated procedure should be executed after all `%test` methods in the current suite. |
+| `%beforeall` | Procedure | Denotes that the annotated procedure should be executed once before all elements of the current suite. |
+| `%afterall` | Procedure | Denotes that the annotated procedure should be executed once after all elements of the current suite. |
 | `%beforeeach` | Procedure | Denotes that the annotated procedure should be executed before each `%test` method in the current suite. |
 | `%aftereach` | Procedure | Denotes that the annotated procedure should be executed after each `%test` method in the current suite. |
 | `%beforetest(<procedure_name>)` | Procedure | Denotes that mentioned procedure should be executed before the annotated `%test` procedure. |
--- a/docs/userguide/annotations.md
+++ b/docs/userguide/annotations.md
@@ -1,34 +1,37 @@
 # Annotations
 
 Annotations provide a way to configure tests and suites in a declarative way similar to modern OOP languages.
+The annotation list is based on jUnit 5 implementation.
 
 # Example
 Let's say we have the test package like this:
 ```
 create or replace package test_pkg is
 
-  -- %suite(Name of suite)
-  -- %suitepackage(all.globaltests)
-  -- %suitetype(critical)
+  -- %suite
+  -- %displayname(Name of suite)
+  -- %suitepath(all.globaltests)
 
-  -- %suitesetup
+  -- %beforeall
   procedure globalsetup;
 
-  -- %suiteteardown
+  -- %afterall
   procedure global_teardown;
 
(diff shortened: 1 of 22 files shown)


docs/userguide/annotations.md
```diff
@@ -1,65 +1,68 @@
...
-%testteardown | Marks that special teardown procedure has to be run before the test instead of the default one
-%suitesetup | Procedure with executes once at the beginning of the suite and doesn't executes before each test
-%suiteteardown | Procedure with executes once after the execution of the last test in the suite.
+| Annotation |Level| Describtion |
+| --- | --- | --- |
+| `%suite` | Package | Marks package to be a suite of tests This way all testing packages might be found in a schema. |
+| `%suitepath(<path>)` | Package | Similar to java package. The annotation allows logical grouping of suites into hierarcies. |
+| `%displayname(<description>)` | Package/procedure | Human-familiar describtion of the suite/test. `%displayname(Name of the suite/test)` |
+| `%test` | Procedure | Denotes that a method is a test method. |
+| `%beforeall` | Procedure | Denotes that the annotated procedure should be executed before all `%test` methods in the current suite. |
```
jgebal: Denotes that the annotated procedure should be executed **before all elements of the current suite**.
If `%beforeall` is defined more then once, the last `%beforeall` in the suite is used.

docs/userguide/annotations.md
```diff
@@ -1,65 +1,68 @@
...
+| `%afterall` | Procedure | Denotes that the annotated procedure should be executed after all `%test` methods in the current suite. |
+| `%beforeeach` | Procedure | Denotes that the annotated procedure should be executed before each `%test` method in the current suite. |
```
jgebal: Each element of suite. We consider sub-suites on a suitepath. If we have suitepath with parent suite package and child suite package, the `%beforeeach` of parent suite will fire before invoking the child suite.

jgebal:
I also like the idea of keeping the annotations as simple as possible.
I'd also like to come back to the `%context` or `%subsuite` idea, where we can nest structures like this:
```
create or replace package demo_equal_matcher as

  -- %suite(Equal matcher)
  -- %suitepackage(org.utplsql.v3.demo.matchers)

  -- %context
  -- %displayname(Comparing user defined types for equality)

    -- %test(Gives success when comparing identical objects containing identical data)
    procedure object_compare_success;

    -- %test(Gives failure when comparing to a null actual)
    procedure object_compare_null_actual;

    -- %test(Gives failure when comparing to a null expected)
    procedure object_compare_null_expected;

    -- %test(Gives success when comparing null actual to a null expected)
    procedure object_compare_null_both_ok;

    -- %test(Gives failure when comparing null actual to a null expected, setting null equal to false)
    procedure object_compare_null_both_fail;

    -- %test(Gives failure when comparing identical objects containing different data)
    procedure object_compare_different_data;

    -- %test(Gives failure when comparing different objects containing identical data)
    procedure object_compare_different_type;

  -- %end_context

end;
/
```

I think that `%suitepath` annotation combined with `%context`/`%endcontext` give huge flexibility and incredible power to the framework.

jgebal:
As we are on the renaming of annotations.
Should we keep the annotations like:
`%suitepath` or rather name it `%suite_path`
Also, should annotation names be case sensitive or not? I think not, as PLSQL is not case-sensitive.
Regardless of the decision on case-sensitivity, we need to explicitly state this in the documentation.

docs/userguide/annotations.md
Pazus: I'm not sure it is what we want... I'd stay with beforeeach/aftereach executes only for tests.

docs/userguide/annotations.md
Pazus: agree

Pazus:
1. I've implemented not all annotations from the list of #133 , Nested and tags will be ready in 3.1
2. I like `suitepath` more, it's less error prone in typing as we dont have compilation spell checker. @rlove, @Shoelace what do you think?
3. annotations are case-insensitive

jgebal:
Can you add all that info to the readme so we have no doubts in the future?
- Annotations are case insensitive.
- If an annotation is defined more then once, the last one in scope of item (suite/test) is used.

Shoelace:
in regards to  %suitepath or %suite_path   i would use not use the underscore and have it as %suitepath

jgebal:
@pazus, can you put strings 'suite' etc. into constants?

PR #146, opened 2017-01-09
Added XUnit reporter

### LINKED ISSUE #60: Implement jUnit XML formatted report
Implement jUnit XML report by creating new reporter class.

Reporter should be able to print it's content to serveroutput, or return as clob/xmltype variable. Most likely the reporter would be configured and processed by a CI using SQlplus so spooling might be a way it can workout.

jgebal:
Resolves #60
Added additional object to store test execution totals.


Pazus:
Also ut_documentation_reporter should report statistics based on `ut_results_counter` racher then gather statistics by itself

source/core/types/ut_test.tpb
```diff
@@ -65,16 +65,21 @@ create or replace type body ut_test as
...
     end if;
     return l_completed_without_errors;
   end;
 
+  overriding member procedure calc_execution_result(self in out nocopy ut_test) is
+  begin
+    self.result := ut_assert_processor.get_aggregate_asserts_result();
+    --expectation results need to be part of test results
+    self.results := ut_assert_processor.get_asserts_results();
+    self.results_count := ut_results_counter(self.result);
```
jgebal: Good point

source/core/types/ut_suite_item.tpb
```diff
@@ -12,6 +12,7 @@ create or replace type body ut_suite_item as
     self.path := nvl(lower(trim(a_path)), self.object_name);
     self.rollback_type := a_rollback_type;
     self.ignore_flag := ut_utils.boolean_to_int(a_ignore_flag);
+    self.results_count := ut_results_counter();
```
jgebal: the counter on suite level aggregates results from items within the suite. It cannot be reset as we would loose the data.

jgebal:
@Pazus. I've updated the documentation_reporter, so it is now using the counters.
At the moment I do not have any better idea about object structures to hold results at test/suite/run level.
I do feel that it is not perfect and can be done better. Maybe we should do it as a separate PR

source/core/types/ut_suite_item.tpb
Pazus: I mean in `do_execute` of an Item. Suite item executes its `do_execute` only once its invoked by parent item. The same goes for tests.

jgebal: But each test is a new object

PR #148, opened 2017-01-09
Isolated API from CORE.

jgebal:
Renamed matchers to conform with naming conventions for UTPLSQL (UT_ prefix).
Added synonyms for matchers, to be exposed as public API.

This PR is a step towards isolating API from the internal UTPLSQL stuff.

With this isolation it's clear, what needs to be exposed to the end-user.
After merging #146 we should move reporters folder into api/reporters or create synonyms for reporters too.


source/install.sql
```diff
@@ -149,31 +149,44 @@ whenever oserror exit failure rollback
...
 prompt Validating installation
--- erors only. ignore warnings
 select name, type, sequence, line, position, text
- from user_errors
-where name not like 'BIN$%'  --not recycled
-and (name like 'UT%' or name in ('BE_FALSE','BE_LIKE','BE_NOT_NULL','BE_NULL','BE_TRUE','EQUAL','MATCH','BE_BETWEEN')) -- utplsql objects
-and attribute = 'ERROR'
+  from user_errors
+ where name not like 'BIN$%'  --not recycled
+   and (name like 'UT%')
```
Shoelace: i think this should probably be ``` name like 'UT\_%' escape '\'```
same as below

Shoelace: hmm actually that would skip the UT package.. so never mind

Pazus: or change to `(name = 'UT' or name like 'UT\_%' escape '\')`

Pazus:
I'd propose to think again about renaming `be_%` to `ut_be%`.
Cons:
1. we reserve both names, be* and ut_be*, user will need to decide all the time if tests are located in the utplsql schema
2. granting. Will neet not only to grant user-accessable object to another schema but also create synonyms.

Hm. Now I think it's not that bad. Anyway user might want to create synonyms as `ut3.ut.expect` is not a convinient way of call.

We need a script creating all grant for a targer user and another for creating appropriate synonyms.

btw. If all framework objects now follow convention we can exclude them schema scan in `ut_suite_manager`.

jgebal:
With this change, all objects are internally referenced by names `UT_...` while the calls from unit tests will be made by synonyms.
This will be more visible when we add scripts to expose API to other schema.
This PR is not adding support for unit tests in other schema, it is only a preparation.
With this change, all the example tests in `/examles` use synonyms for marchers.

source/install.sql
jgebal: That change would be nice

PR #152, opened 2017-01-11
Run as other user

Pazus:
Implemented ability to tun ut3 from another schema
added fail procedure to `ut` interface
added global comparison null behavior control to `ut` interface
reorganized examples, divided them for user and developer
Only User examples are executed by not-owner user
Fixed bug when `ut_runner` failed with error when there is no tests in provided schema


Shoelace:
this looks like a better version then mine..  plus #112 is quite out of date now..
thanks for taking the time.. i havent had much lately.

Shoelace:
I think your synonyms are wrong..
```create or replace synonym &ut3_user .be_between for &ut3_owner .be_between;```
should be somethign more like
```create or replace synonym &ut3_user .be_between for &ut3_owner .ut_be_between;```

yes?

**Comment by Shoelace** [comment]
no.. never mind..  on my system eth owner does not have create synonym privilege  so the user sysnonyms where pointing to nothing

**Comment by jgebal** [comment]
@shoelace, the optionality of dbms_pipe will be addressed separately. It will go together with dbms_output buffer

**Comment by Shoelace** [comment]
cool  thanks @jgebal .  the only other comment on this i'll make then would be to suggest splitting client_source\grant_user.sql into separate grant and synonyms scripts..

File: .travis/create_utplsql_owner.sql [hunk 1]
```diff
@@ -0,0 +1,24 @@
...
+define ut3_user       = &1
+define ut3_password   = &2
+define ut3_tablespace = &3
+
+create user &ut3_user identified by &ut3_password default tablespace &ut3_tablespace quota unlimited on &ut3_tablespace;
+
+grant create session, create procedure, create type, create table, create synonym to &ut3_user;
+
+grant execute on dbms_pipe to &ut3_user;
+grant create job to &ut3_user;
```
jgebal: This grant is not needed at the moment, can we remove it? I've added it with idea of using dbms_scheduler, but that idea is dropped for now.

examples/demo_expectations.pck
```diff
@@ -85,15 +85,15 @@ create or replace package body demo_expectations is
     ut.expect( to_char(null), 'this should fail' ).not_to( equal( to_char(null) ) );
 
     --fails on nulls not beeig equal
-    ut_assert_processor.nulls_are_equal(false);
+    ut.nulls_are_equal(false);
```
jgebal: Hmm...
This procedure was actually meant to become an annotation.
It would be much better if we could keep the `null_are_equal` contained in a way we have it with transaction handling etc.
So that when you annotate a suite or test with
`-- %nulls_equal`
The scope of the change is on a whole test or whole suite level.

I think the null_are_equal as a procedure should not be exposed to users because, when you set it in one tests it affects other tests unless you reset it back to default.

If we would agree to have such annotation then for this PR I'd suggest one of two things:
- comment the tests that use that functionality until we have annotation
- add new annotation support as part of this PR

I guess the firs one seems more appropriate, so we do not have too many changes in a PR.

What do you think?

source/api/ut.pkb
```diff
@@ -64,6 +64,21 @@ create or replace package body ut is
   begin
     return ut_expectation_dsinterval(ut_data_value_dsinterval(a_actual), a_message);
   end;
+  
+  procedure fail(a_message in varchar2) is
```
jgebal: Can you comment on why are we exposing this to the user?
I don't understand the background for this being used by user.

Pazus: sometimes it familiar for user to set a point in code, if it's riched than something is wrong. The most frequent case (but not the only) is when exception is expceted.

Other frameworks also have such methods (http://junit.sourceforge.net/javadoc/org/junit/Assert.html)

examples/demo_expectations.pck
Pazus: I agree that current configuration is not good. I remove it from ut. We'll decide how to do it more approprietly later. I'm glad we've tried it, discussed and made intelligent decision.

source/api/ut.pkb
jgebal: Fair point.
I'm wondering how the reporter will show the line of failure. I'm afraid it might be getting it wrong.

jgebal:
We're missing ability to grant to public and create public synonyms.
In my projects we use a lot of schemes. Having public grants and synonyms is much more convenient than granting it to each schema individually.

source/api/ut.pkb
Pazus: Fair point!

jgebal: In this case, it doesn't report the line - good.

PR #154, opened 2017-01-12
Restored old syntax

Pazus:
Restored old syntax as a way to define test/suite name additional to `%displayname`
If defined both `%displayname` is used


jgebal:
Can you also update the annotations documentation?

PR #156, opened 2017-01-14
Separate logical suites and real one

Pazus:
Introduced new object type - `ut_real_suite` which is contructed for anntoated packages. For logical suites stripped `ut_suite` type is used which does nothing but grouping real suites.

`ut_suite` doesn't issue savepoint/rollback.


jgebal:
I think it's a change but I don't really like the proposed name `ut_real_suite`
I have 2 alternative names to suggest:
- `ut_sute` is the real suite
- `ut_logical_suite` is the base for `ut_suite`
or:
- `ut_suite_package` is the real suite
- `ut_logical_suite` is the base for `ut_suite_package`
I personally prefer the first option.

Pazus:
renamed object types

PR #159, opened 2017-01-16
Added colors to the documentation reporter.

jgebal:
Added option to ut_run to enable colors.
Added colors to Travis build.


jgebal:
Need to fix indents on documentation reporter.
Need to review the overall format of failures reporting.

jgebal:
For colors to show on Windows 7, you need ansicon installed:
https://github.com/adoxa/ansicon/releases/tag/v1.66

source/core/types/ut_reporter_base.tps
```diff
@@ -5,6 +5,21 @@ create or replace type ut_reporter_base force authid current_user as object
   constructor function ut_reporter_base(self in out nocopy ut_reporter_base, a_output ut_output default ut_output_dbms_output()) return self as result,
 
   member procedure print_text(self in out nocopy ut_reporter_base, a_text varchar2),
+
```
Pazus: I still don't think this should be base object type methods.

source/core/ut_color_helper.pks
```diff
@@ -0,0 +1,11 @@
+create or replace package ut_color_helper as
```
Pazus: I propose to rename it to `ut_console_color_helper` or `ut_ansiconsole_coloring_helper` to proper reflect it's functionality

jgebal: how about `ut_ansiconsole_helper`?

source/core/types/ut_reporter_base.tps
jgebal: OK, they are console specific.
I could create another abstract on top of `ut_reporter` or just move it to `ut_documentation_reporter`.
Thinking about making the API extensible, maybe another abstract would be good.
Maybe a name like `ut_console_reporter` to keep console specific API?

source/core/ut_color_helper.pks
Pazus: good

source/core/types/ut_reporter_base.tps
Pazus: Ok or simply put it into `ut_documentation_reporter`. If you feel it's going to be reused, abstraction is ok for me as well.

Pazus:
Need tp think but I liked the previous way of passing parameter more...

Pazus:
I mean its not the runner's param, it the reporter's attributed to be set before the reporter is passed to the runner.

jgebal:
For the color param. I easy to keep API super-simple. Better to have one-liner to run the tests:
`exec ut_runner.run(...)` than do multiline setup before execution.
Color seems to me as part of reporting for console (with potentially different reporter). If sql developer will support it (I'll ask for it) then it is a valid use for developers).

Pazus:
`exec ut_runnes.run(user,ut_documentation_reporter(true));` ?

And the commandline script remains the same.

jgebal:
It is close but not the same.
`ut_runner.run(user,true);`
`ut_runner.run(user,ut_reporters(...,...,...),true);`
It also makes this one reporter "special" as it has different constructor than others.

Pazus:
But i'm sure in the future reporters will have parameters. Not alot, but they will, and such parameters will be very specific to the individual reporter

jgebal:
Here is the thing.
In proposed implementation, the ut_documentation_reporter does not care about the `a_color_console` parameter.
It is not even aware of it's existence.
This is because the `ut_console_reporter_base` is taking care of it.
So any reporter can use calls to:
- `print_red_text`
- `print_green_text`
- `print_yellow_text`
- `print_blue_text`
- `print_cyan_text`
- `print_magenta_text`
It is the core of framework that makes decision if colors should be used or not.
If we were to go with approach of setting it on a reporter level, than every "text" reporter would need to have it's own implementation for it and make decision on his own.
Going towards this direction would also mean we should get rid of the `ut_console_reporter_base` as an abstraction since it doesn't bring any value.

Pazus:
Have you found a way to define if coloring can be used?

jgebal:
I've gogoled 2 ways:
http://unix.stackexchange.com/questions/9957/how-to-check-if-bash-can-print-colors
http://stackoverflow.com/questions/7445658/how-to-detect-if-the-console-does-support-ansi-escape-codes-in-python

I don't think it's doable in sqlplus :)

jgebal:
Please hold off with merging - need to fix the ut_run.sql when running as other user.

jgebal:
Issues are now resolved and we're good for merging

Issue #134, opened 2016-12-23
ut.runner failes with wrong schema provided


Pazus:
it failes with not nice exception
```
begin
  ut.run('123123',ut_documentation_reporter);
end;
```

jgebal:
I think this is a valid behaviour. The only thing is that framework should throw a named exception like: "invalid schema name (the schema name) provided" / "Invalid package name (the package name) provided" / "Suitepath (the path) not found"
The same goes when you specify valid schema, but non existing package.

I wonder what would happen if I specify a valid schema and a valid package, but the package is not a utPLSQL v3 package.

Issue #161, opened 2017-01-16
calling ut_output_pipe_helper.purge


Pazus:
in the `ut_runner` with NDS `ut_output_pipe_helper.purge` is called but it doesn't exist

PR #162, opened 2017-01-16
Made ut_runner to fail on invalid path

Pazus:
Made ut_runner to fail if non-existent schema is provided or provided package is not a suite package.

Refers #134


source/core/ut_suite_manager.pkb
```diff
@@ -411,6 +411,8 @@ create or replace package body ut_suite_manager is
     for i in 1 .. l_paths.count loop
       l_path   := l_paths(i);
       l_schema := regexp_substr(l_path, '^(\w+)(\.|:|$)', 1, 1, null, 1);
+      
+      l_schema := sys.dbms_assert.schema_name(upper(l_schema));
```
jgebal: Nice, I didnt use dbms_assert before. Good to know about it's existence.

source/core/ut_suite_manager.pkb
```diff
@@ -434,6 +436,11 @@ create or replace package body ut_suite_manager is
             l_procedure_name := regexp_substr(l_path, '^\w+\.(\w+)(\.(\w+))?$', subexpression => 3);
 
             l_temp_suite := config_package(l_schema, l_package_name);
+
```
jgebal: Maybe use dbms_assert.sql_object_name first, and if it fails, we know the object name is not valid.
Then do the if statement, and raise error stating, that the package is not a properly annotated unit test suite package.

Pazus:
I've tested it manually but its not clear to write automated tests for that as we still have problem with `purge` method dynamic call which doesn't exists

Pazus:
It failes on wrong schema, it failes on non-existent package and if provided package is not a suite package

Pazus:
#161

jgebal:
Please remove the purge from `run`. I will add it back properly implemented with changes to outputs.

jgebal:
Never mind. I'll be making changes to this code anyway.

PR #166, opened 2017-01-17
Create release archive

rlove:
These changes:

- Build documentation HTML
- Creates release directory with files organized for release
- Creates archives based on release directory
    - utplsql.zip 
    - utplsql.tar.gz

- Currently the following directories are not deployed  (But easily changed)
   - .git
   - .travis
   - development
   - client_source
   - tests

NOTE:
This does not automatically create a release in github.   
I am separating that into a different pull request, as this one is easy to verify it works in travis.


rlove:
Forced Merged Pull request as it needs to be implemented to be able to to the release today.

PR #174, opened 2017-01-19
Feature/dbms output buffer

jgebal:
This change addresses the issue of using more than one reporter with dbms_output.
It is a complete change though not a complete feature.
With this change, outputs from each reporter are buffered in a buffer build as a temporary clob.
At the end, outputs are displayed from each reporter.
Once outputs rework is complete it will bring more value as we will also be able to redirect each output buffer (like we can do with pipelined output).
The differences in how it works are visible when looking at composite reporter example:
[new behavior](https://travis-ci.org/jgebal/utPLSQL#L2247-L2271)
[old behavior](https://travis-ci.org/jgebal/utPLSQL#L2247-L2271)


source/core/ut_output_buffer.pkb
```diff
@@ -0,0 +1,64 @@
...
+  begin
+    if a_text is not null then
+      add_to_buffer(a_output_id, a_text||chr(10));
+    end if;
+  end;
+
+  procedure add_to_buffer(a_output_id varchar2, a_text clob) is
+    l_clob clob;
+  begin
+    if a_text is not null then
```
Pazus: and `dbms_lob.getlength(a_text)>0` as after `dbms_lob.createtemporary` clob variable is not null though the length is 0.

jgebal:
I'm cancelling this PR for now - new approach will be taken.

PR #176, opened 2017-01-21
renamed "disable" annotation to "disabled"

Pazus:
Renamed `disable` annotation to `disabled`


jgebal:
Nice change.

jgebal:
I think we should use label `bug` for issues and `bugfix` for pull requests
.

PR #177, opened 2017-01-21
Cleanup of outputs - removing pipe and dbms_output as a buffer

jgebal:
We now have only one kind of output buffer (table).
This simplifies design significantly.
The table has no constraints on size (like in-memmory buffers)
The table allows cross-session communication.
Implemented buffering should keep the table small, if consumers use it as they should.
Any garbage that is older than 1 day will be removed on every run of utPLSQL.


source/core/ut_output_buffer.pkb
```diff
@@ -0,0 +1,100 @@
...
+    loop
+      delete from (
+        select *
+          from ut_output_buffer_tmp where reporter_id = a_reporter_id order by message_id
+        )
+      returning text bulk collect into l_results;
+
+      --nothing fetched from output, wait and try again
+      if l_results.count = 0 then
+        dbms_lock.sleep(c_sleep_time);
```
jgebal: We need to seep for a short while to avoid bombarding database with delete statements when zero rows produced by the run.

jgebal: **I'm having a second thought.**
Maybe we should use `dbms_pipe` for signalling between sessions.
That would be optional, and if users don't want to use it, they get the option with `dbms_lock`

**How signalling with `dbms_pipe` would work?**
We would create one pipe for one reporter instance.
After saving data to buffer reporter sends signal, that new data has arrived.
Consumer (`read_lines`), first waits on a pipe_message, and when received, reads (and deletes) all data from buffer.
Messages would be very lightweight - it is just a ping to wakeup the process.

So the clients installing utPLSQL would need to decide. Either they use `dbms_pipe` and have lightweight signalling (push notification to consumer).
Or they don't want to use it, and then the `read_lines` uses a pull mechanism using `dbms_lock` to minimise the frequency of SQL statements being issued.

The install could have option to choose (two separate install scripts?).
We would use conditional compilation in `ut_output_buffer.get_lines` to decide if `dbms_pipe` or `dbms_lock` is used.

source/core/ut_output_buffer_tmp.sql
```diff
@@ -0,0 +1,17 @@
+create table ut_output_buffer_tmp(
+  /*
+  * This table is not a global temporary table as it needs to allow cross-session data exchange
+  * It is used however as a temporary table with multiple writers.
+  * This is why it has very high initrans and has nologging
+  */
+  reporter_id    raw(32) not null,
+  message_id     number(38,0) not null,
+  text           varchar2(4000),
```
Shoelace: minor quibble.. should you not have the nullable "long varchar" as the last column? that shoudl help with space usage..

Pazus: Does it matte?

Shoelace:
not exactly,  but its always good to follow good database design.


--- David


________________________________

PR #189, opened 2017-01-31
Feature/coverage

jgebal:


source/core/coverage/ut_coverage.pkb
```diff
@@ -0,0 +1,135 @@
...
+    l_new_unit         t_unit_coverage;
+    l_skipped_objects  ut_object_names := ut_object_names();
+  begin
+
+    if not ut_coverage_helper.is_develop_mode() then
+      l_skipped_objects := ut_utils.get_utplsql_objects_list() multiset union set(g_skipped_objects);
+    end if;
+
+    -- TODO - add inclusive and exclusive filtering
+    select ut3.ut_coverage_row(
```
Shoelace: you need to remove the ut3 references...

jgebal: Thanks. Good catch

Pazus:
Great work @jgebal . Nice clean code.

source/core/coverage/ut_coverage.pkb
```diff
@@ -0,0 +1,135 @@
...
+      left join table( ut3.ut_coverage_helper.get_raw_coverage_data() ) u
+        on s.owner = u.owner
+       and s.name  = u.name
+       and s.type  = u.type
+       and s.line  = nvl(u.line_number, s.line)
+     where s.type not in ('PACKAGE', 'TYPE')
+       and s.owner in (select t.column_value from table(g_schema_names) t)
+       --Exclude calls to utPLSQL framework and Unit Test packages
+       and not exists(select 1 from table(l_skipped_objects) l where s.owner = l.owner AND s.name = l.name)
+     order by s.owner, s.name, s.line;
```
jgebal: Some help would be appreciated on this query. It is very slow but the main reason is usage of all_source instead of dba_source.
Maybe we need to revert to checking if dba source is accessible?

Pazus: try this one:
```
  select /*+dynamic_sampling(t 2)*/ut3.ut_coverage_row(
             lower(s.owner||'.'||s.name),
             s.owner, s.name, s.type, s.line,
           --filtering out false - negatives reported by profiler (zero executions while it should be ignored).
           case when
              regexp_like(
                s.text,
                '^\s*(((not)?\s*(overriding|final|instantiable)\s*)*(constructor|member)?\s*(procedure|function)|begin|end\s*;)', 'i'
              ) then u.total_occur
           end
          )
      bulk collect into l_data
      from all_source s
      join (select distinct column_value owner from table(cast (g_schema_names as ut_varchar2_list))) t
        on t.owner = s.owner
      --need to use owner privileges here
      left join table( ut3.ut_coverage_helper.get_raw_coverage_data() ) u
        on s.owner = u.owner
       and s.name  = u.name
       and s.type  = u.type
       and s.line  = nvl(u.line_number, s.line)
     where s.type not in ('PACKAGE', 'TYPE')
       --Exclude calls to utPLSQL framework and Unit Test packages
       and (s.owner, s.name) not in (select /*+dynamic_sampling(l 2)*/ l.owner, l.name from table(cast(l_skipped_objects as ut_object_names)) l)
     order by s.owner, s.name, s.line;
```

source/core/coverage/proftab_temp_tables.sql
```diff
@@ -0,0 +1,59 @@
+create global temporary table plsql_profiler_runs
+(
+  runid           number primary key,  -- unique run identifier,
+                                       -- from plsql_profiler_runnumber
+  related_run     number,              -- runid of related run (for client/
+                                       --     server correlation)
+  run_owner       varchar2(32),        -- user who started run
```
jgebal: CC?

source/core/coverage/ut_coverage.pks
```diff
@@ -0,0 +1,54 @@
...
+    covered_lines   binary_integer := 0,
+    uncovered_lines binary_integer := 0,
+    total_lines     binary_integer := 0,
+    executions      number(38,0)   := 0,
+    objects         tt_program_units
+  );
+
+  function  get_coverage_id return integer;
+
+  function  coverage_start(a_schema_names ut_varchar2_list := ut_varchar2_list(sys_context('userenv','current_schema'))) return integer;
```
jgebal: This might have been true on Oracle 10. With 11g XE it works fine.
I've tested the provided example and go the following:
```sql
conn ut3/ut3
create or replace procedure definer_proc is
  begin
    dbms_output.PUT_LINE('Current user=' || sys_context('userenv', 'current_user'));
    dbms_output.PUT_LINE('Session user=' || sys_context('userenv', 'session_user'));
    dbms_output.PUT_LINE('Current schema=' || sys_context('userenv', 'current_schema'));
  end definer_proc;
/
create or replace procedure invoker_proc authid current_user is
  begin
    dbms_output.PUT_LINE('Current user=' || sys_context('userenv', 'current_user'));
    dbms_output.PUT_LINE('Session user=' || sys_context('userenv', 'session_user'));
    dbms_output.PUT_LINE('Current schema=' || sys_context('userenv', 'current_schema'));
  end invoker_proc;
/
grant execute on definer_proc to ut3_user;
grant execute on invoker_proc to ut3_user;
begin ut3.definer_proc; end;
/
begin ut3.invoker_proc; end;
/

sql> begin ut3.definer_proc; end;
Current user=UT3
Session user=UT3
Current schema=UT3
sql> begin ut3.invoker_proc; end;
Current user=UT3
Session user=UT3
Current schema=UT3
```

```sql
conn ut3_user/ut3_user
begin ut3.definer_proc; end;
/
begin ut3.invoker_proc; end;
/  
sql> begin ut3.definer_proc; end;
Current user=UT3
Session user=UT3_USER
Current schema=UT3
sql> begin ut3.invoker_proc; end;
Current user=UT3_USER
Session user=UT3_USER
Current schema=UT3_USER

```

source/core/coverage/ut_coverage.pkb
```diff
@@ -0,0 +1,135 @@
...
+              regexp_instr(
+                s.text,
+                '^\s*(((not)?\s*(overriding|final|instantiable)\s*)*(constructor|member)?\s*(procedure|function)|begin|end\s*;)', 1, 1, 0, 'i'
+              ) = 0 then u.total_occur
+           end
+          )
+      bulk collect into l_data
+      from all_source s
+      --need to use owner privileges here
+      left join table( ut3.ut_coverage_helper.get_raw_coverage_data() ) u
```
jgebal: It's a question of:
- do we do it in one big shot or do we want to do it row by row (object by object)?
- object by object might generate many individual queries to dba_source - but this actually might be faster than one big join ans as you said - there is a risg of large data volumes getting generated.

source/core/coverage/ut_coverage_helper.pkb
```diff
@@ -0,0 +1,93 @@
...
+         and u.unit_type not in ('PACKAGE SPEC', 'TYPE SPEC', 'ANONYMOUS BLOCK');
+  begin
+    open l_coverage_data( get_coverage_id());
+    loop
+      fetch l_coverage_data bulk collect into l_results limit 1000;
+      for i in 1 .. l_results.count loop
+        pipe row (l_results(i));
+      end loop;
+
+      exit when l_coverage_data%notfound;
```
jgebal: This is why I'm doing exit when AFTER looping through results :)
When last batch is returned by bulk collect into the cursor variable will indicate %NOTFOUND but the collection might contain between 0 and `limit-1` rows.
So I loop through the collection and exit after that.

source/core/coverage/ut_coverage_helper.pkb
```diff
@@ -15,6 +15,10 @@ create or replace package body ut_coverage_helper is
 
   function coverage_start(a_run_comment varchar2) return integer is
   begin
+    --those are Global Temporary tables for profiler usage only
+    delete from plsql_profiler_data;
```
Pazus: Do these tables have public delete grants?

jgebal: `ut_coverage_helper` runs with authid definer.
The main reason for separating `ut_coverage_helper` out of `ut_coverage` was difference in authid between those two. `ut_coverage_helper` isolates users from underlying objects so you can change `ut_coverage_helper` to use 12 R2 "coverage" or something else (in theory) and don't need to change anything else than this package.

source/core/coverage/ut_coverage.pkb
```diff
@@ -65,68 +65,104 @@ create or replace package body ut_coverage is
...
+    select s.owner,s.name,s.line,s.text
       from all_source s
-      --need to use owner privileges here
-      left join table( ut_coverage_helper.get_raw_coverage_data() ) u
-        on s.owner = u.owner
-       and s.name  = u.name
-       and s.type  = u.type
-       and s.line  = nvl(u.line_number, s.line)
      where s.type not in ('PACKAGE', 'TYPE')
        and s.owner in (select t.column_value from table(g_schema_names) t)
```
Pazus: I recommend to put /*+dynamic_sampling(t 2)*/ for optimized to start with the size of collection, or we can set /*+cardinality(5)*/ which is not as good to mark that the list of schemas is small

source/core/coverage/ut_coverage.pkb
```diff
@@ -65,68 +65,104 @@ create or replace package body ut_coverage is
...
+
+      --if there is coverage, we need to filter out the garbage (badly indicated data from dbms_profiler)
+      if l_line_calls.count > 0 then
+        --get source lines to skip
+        select line
+          bulk collect into l_source_lines
+          from ut_coverage_sources_tmp c
+         where c.owner = src_object.owner
+           and c.name = src_object.object_name
+           and regexp_instr(
```
Pazus: Do you think we should filter out while populating `ut_coverage_sources_tmp`?
I think you the use of `regexp_like` is more clear then `regexp_instr() != 0`

jgebal:
Sorry for so many commits. I think this code is now ready for merging.

source/core/coverage/ut_coverage.pkb
jgebal: Done.

Pazus:
Could you please actualize the branch

jgebal:
Will update it today.

PR #196, opened 2017-02-04
Suitepath current user shortcut

Pazus:
Added support for suitepath sortcutt for current schema.
As from update doc file:
A `%suitepath` can be provided in tree ways:
* schema - execute all test in the schema
* [schema]:suite1[.suite2][.suite3]... - execute all tests in all suites from suite1[.suite2][.suite3] path. If schema is not provided, then current schema is used. Example: `:all.rooms_tests`.
* [schema.]package[.procedure] - execute all tests in the test package provided. The whole hierarchy of suites in the schema is build before, all before/after hooks of partn suites for th provided suite package are executed as well. Example: `tests.test_contact.test_last_name_validator` or simply `test_contact.test_last_name_validator` if `tests` is the current schema.


jgebal:
Nice work. I've only updated the readme, as `[.procedure]` is acceptable for both pakcage ans suitepath syntax.

PR #202, opened 2017-02-26
Hided output table under view to be EBR-ready

Pazus:
This is needed to be EBR ready. A user might want to test new version of the framework in another edition to not break current state. We have to support that.

### DIFF
--- a/source/core/ut_output_buffer_tmp.sql
+++ b/source/core/ut_output_buffer_tmp.sql
@@ -1,4 +1,4 @@
-create table ut_output_buffer_tmp(
+create table ut_output_buffer_tmp$(
   /*
   utPLSQL - Version X.X.X.X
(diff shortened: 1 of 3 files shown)


source/core/ut_output_buffer_tmp.sql
```diff
@@ -27,4 +27,37 @@ create table ut_output_buffer_tmp(
...
+  pragma exception_init(ex_view_doesnt_exist,-942);
+begin
+  begin
+    execute immediate 'drop view ut_output_buffer_tmp';
+  exception
+    when ex_view_doesnt_exist then
+      null;
+  end;
+  
+  execute immediate 'create or replace editioning view ut_output_buffer_tmp as
```
jgebal: Please add the copyright banner to the views too.

PR #205, opened 2017-02-26
Added outputs to the reporters. Moved *each to ut_test

Pazus:
Now `ut_executable` stores the stderr, stdout of the execution wich can be used by reporters.
`ut_documentations_reporter` prints the output of its steps along with test results (in future we might disable that by parameter)
`ut_teamcity_reporter` also reports output and stderr according to the Teamcity messaging format.
Added new tests to check new behavior

Moved beforeeach/aftereach executables from `ut_suite` to `ut_test`, fixed the behavior on failure.

### DIFF
--- a/source/core/ut_suite_manager.pkb
+++ b/source/core/ut_suite_manager.pkb
@@ -120,9 +120,7 @@ create or replace package body ut_suite_manager is
           a_rollback_type         => l_suite_rollback,
           a_ignore_flag           => l_annotation_data.package_annotations.exists('disabled'),
           a_before_all_proc_name  => l_suite_setup_proc,
-          a_after_all_proc_name   => l_suite_teardown_proc,
-          a_before_each_proc_name => l_default_setup_proc,
-          a_after_each_proc_name  => l_default_teardown_proc
+          a_after_all_proc_name   => l_suite_teardown_proc
       );
 
 
@@ -174,7 +172,9 @@ create or replace package body ut_suite_manager is
                 a_rollback_type => l_rollback_type,
                 a_ignore_flag   => l_proc_annotations.exists('disabled'),
                 a_before_test_proc_name => l_setup_procedure,
-                a_after_test_proc_name  => l_teardown_procedure
+                a_after_test_proc_name  => l_teardown_procedure,
+                a_before_each_proc_name => l_default_setup_proc,
+                a_after_each_proc_name  => l_default_teardown_proc
             );
 
             l_suite.add_item(l_test);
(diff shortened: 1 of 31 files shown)


jgebal:
@pazus, 
Here is an idea on how to handle before/after calls as a list.

```sql
l_before_each_list := ut_executables(
  ut_executable('the before each'),
  ut_executable(null) -- no before test defined
);
l_after_each_list := ut_executables(
  ut_executable('the after test'),
  ut_executable(null) -- no after test
);

--execute the before blocks
i := l_before_each_list.first;
while i is not null loop
  if l_before_each_list(i) is not null then 
    l_before_each_list(i).do_execute;
  end if;
  l_before_exec_count := l_before_exec_count + 1;
  i := l_before_each_list.next(i);
end loop;

-- then you track the count of executed before block and execute
-- the after blocks in following loop
i := l_after_each_list.last;
while i is not null loop
  if l_after_each_list(i) is not null then 
    l_after_each_list(i).do_execute;
  end if;
  l_after_exec_count := l_after_exec_count + 1;
  exit when l_after_exec_count >= l_before_exec_count;
  i := l_after_exec_count .prior(i);
end loop;
```

That is a rough idea - not a real implementation, but this would also open doors for propagating `beforeeach` from parents (parent suites or within suite when we introduce sub-suites/contexts).

What do you think about this?

Pazus:
Should I implement it in the current PR?

source/reporters/ut_documentation_reporter.tpb
```diff
@@ -44,6 +44,17 @@ create or replace type body ut_documentation_reporter is
 
   overriding member procedure after_calling_test(self in out nocopy ut_documentation_reporter, a_test ut_test) as
     l_message varchar2(4000);
+    
+    procedure print_output(a_exectable ut_executable) is
+      l_lines ut_varchar2_list;
+    begin
+      if a_exectable is not null and a_exectable.is_defined and a_exectable.serveroutput is not null and dbms_lob.getlength(a_exectable.serveroutput) > 0 then
```
jgebal: typo in var name

source/core/ut_suite_manager.pkb
```diff
@@ -174,7 +172,9 @@ create or replace package body ut_suite_manager is
                 a_rollback_type => l_rollback_type,
                 a_ignore_flag   => l_proc_annotations.exists('disabled'),
                 a_before_test_proc_name => l_setup_procedure,
-                a_after_test_proc_name  => l_teardown_procedure
+                a_after_test_proc_name  => l_teardown_procedure,
```
jgebal: can you cleanup the variable names to match new conventions?

source/reporters/ut_documentation_reporter.tpb
```diff
@@ -44,6 +44,17 @@ create or replace type body ut_documentation_reporter is
 
   overriding member procedure after_calling_test(self in out nocopy ut_documentation_reporter, a_test ut_test) as
     l_message varchar2(4000);
+    
+    procedure print_output(a_exectable ut_executable) is
```
jgebal: Can we handle printing outputs of `beforeall` `afterall` as well?

tests/ut_test/ut_test.TestOutputGathering.sql
```diff
@@ -0,0 +1,88 @@
+create or replace package ut_output_tests
+as
+ --%suite
+
```
jgebal: Can we add beforeall/afterall?

Pazus:
@jgebal Thanks for the comments, I've addressed them.

PR #214, opened 2017-03-04
Implemented coveralls reporting

Pazus:


Pazus:
Build is failing cause we need to define COVERALLS_REPO_TOKEN variable in travis-ci

PR #218, opened 2017-03-05
Update of badges in readme

jgebal:


Pazus:
And what about master? For master it will also report on develop branch?

jgebal:
Well, the trouble with coveralls badge is that it is reporting on a specified branch (not on the current branch).
We would need a way to update readme.md with branch name automatically.
I think the most important is to have coverage on the development branch.

PR #226, opened 2017-03-11
Enchance reporting

Pazus:
Fixed ut_run.sql to execute files from working folder
Enhanced Teamcity reporter
Fixed xunit reporter as otherwise it treated passed tests as skipped in Teamcity


source/reporters/ut_teamcity_reporter.tpb
```diff
@@ -24,84 +24,110 @@ create or replace type body ut_teamcity_reporter is
...
       print_output(a_test.item);
       print_output(a_test.after_test);
       print_output(a_test.after_each);
-
-      if a_test.results is not null and a_test.results.count > 0 then
+    
+      if a_test.result = ut_utils.tr_error then
+        self.print_text(ut_teamcity_reporter_helper.test_std_err(a_test_name => l_test_full_name
+                                                                ,a_out       => coalesce(a_test.before_each.error_backtrace
+                                                                                        ,a_test.before_test.error_backtrace
```
jgebal: If both `before_test` and `after_test` will throw, then we will not show the stacktrace from the after block.

PR #232, opened 2017-03-15
Travis Docker Config

viniciusam:
Changes to use private docker images to test against 12c and 11g.


jgebal:
Thanks @viniciusam - great work and huge progress.
Now we have 2 Oracle versions running our framework.
Cant wait for instructions in contriguting.md and 12.2 as a third build option :D

viniciusam:
No problem @jgebal, I am happy to help. I will start to work on that instructions.
About the 12.2 version, It's easy to add but the image is even bigger than 12.1, the build time will be longer...

PR #237, opened 2017-03-25
Cleanup and simplifications to `.travis.yml` file.

jgebal:
Cleanup of scripts in `.travis` directory.
Build performance improvement - minimizing invocations of `sqlcl` as it takes time to start it.


.travis/install.sh
```diff
@@ -1,16 +1,33 @@
...
 #install core of utplsql
-"$SQLCLI" $UT3_OWNER/$UT3_OWNER_PASSWORD@//$CONNECTION_STR @install.sql $UT3_OWNER
+"$SQLCLI" sys/$ORACLE_PWD@//$CONNECTION_STR AS SYSDBA <<-SQL
+pwd
+set feedback off
+set verify off
 
+@../source/create_utplsql_owner.sql $UT3_OWNER $UT3_OWNER_PASSWORD $UT3_TABLESPACE
+--only needed to run unit tests for utplsql v3, not required to run utplsql v3 itself
+grant select any dictionary to $UT3_OWNER;
```
Pazus: Are you sure we need it? I think my used doesn't have it and works fine

.travis/install.sh
```diff
@@ -1,16 +1,33 @@
...
 
+@../source/create_utplsql_owner.sql $UT3_USER $UT3_USER_PASSWORD $UT3_TABLESPACE
```
Pazus: you call the script twice? Seems like a bug

.travis.yml
```diff
@@ -19,51 +19,39 @@ services:
...
-    - ORACLE_VERSION=$ORACLE_11gR2XE
+    - ORACLE_VERSION="12c-se-r1" CONNECTION_STR='127.0.0.1:1521/ORCLPDB1'
+    - ORACLE_VERSION="11g-xe-r2" CONNECTION_STR='127.0.0.1:1521/XE'        DOCKER_OPTIONS='--shm-size=1g'
+
 
 cache:
+  pip: true
   directories:
     - $CACHE_DIR
+    - node_modules
```
viniciusam: You don't need to cache the node_modules dir, since we only use it to download sqlcl. The downloaded file is cached, so next time we won't need to install those dependencies again.

Caching it will only increment the cache fetch time.

viniciusam: Sorry, didn't realize that this is also used for coveralls_uploader.sh.

.travis/install.sh
jgebal: It is called with different parameters. 
I got rid of duplicated scripts in `.travis` directory.
Maybe we should rename script to be: `create_utplsql_user.sql`

jgebal: It is called with different parameters. 
I got rid of duplicated scripts in `.travis` directory.
Maybe we should rename script to be: `create_utplsql_user.sql`

.travis/install.sh
jgebal: It is needed for ut_coverage to work.
That should be fixed/addressed separately.
I just noticed that yesterday.
Also the coverage should use dba_source for best results.

jgebal: I've created a separate issue to keep it in mind.

jgebal:
I will merge this PR. There seems to be issues with cache for this PR and therefore the build is failing. Build was fine on my repo.

PR #248, opened 2017-03-27
Update of the release process to make it update version of project files

jgebal:
Release process is automated in following way:
1) with every build, we update files with appropriate version number before we deploy
2) when build is executed on a branch named `release/v1.2.3-something` then additional steps are taken:
    - project version in files: `sonar-project.properties`, `VERSION` is updated from the version number derived from release branch
    - changes on those two files are committed and and pushed - this should happen only once, when the release branch is initially created on the main repo
3) To create a release, just create a tag on the code to be released. The tag name must match the regex pattern: `^v[0-9]+\.[0-9]+\.[0-9]+.*$`
   - When a tag build is executed, the documentation is built and files are uploaded to the tag.
   - The version number is derived from the tag name.
4) Release version do not provide access to unversioned source files (the default zip file from github is built to be empty). The sources for release are provided in separate zip files delivered from travis build process.


jgebal:
@Pazus 
I've added release notes to development/releasing.md
Can you approve the PR?

jgebal:
We need to resolve conflicts and we should be good.

.travis/build_release_archive.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+mv -f .gitattributes.release .gitattributes
+git add .
```
Pazus: I think we shouldn't add all, we should add specific files

jgebal: The add & commit is used only to make use of `git archive` feature.
The committed changes are never pushed to the repository.

PR #252, opened 2017-03-29
Added support for excluding columns in cursor comparison.

jgebal:
There are two ways to pass column list:
- as a comma separated values
- as a ut_varchar2_list collection


jgebal:
I'm cancelling this PR - it needs to be reworked.

PR #271, opened 2017-04-09
Fixed privileges issues needed for coverage 

jgebal:
Added grant select any dictionary.

Fixed issue with weak password for ut3 user (some databases have password policies that require stronger passwords).
Added grant on DBMS_LOB, some user reported that this was not granted to public on his database.
Updated documentation to reflect changes.

### DIFF
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -9,8 +9,6 @@ set feedback off
 set verify off
 
 @../source/create_utplsql_owner.sql $UT3_OWNER $UT3_OWNER_PASSWORD $UT3_TABLESPACE
---only needed to run unit tests for utplsql v3, not required to run utplsql v3 itself
-grant select any dictionary to $UT3_OWNER;
 
 @../source/create_utplsql_owner.sql $UT3_USER $UT3_USER_PASSWORD $UT3_TABLESPACE
 cd ..
--- a/docs/userguide/install.md
+++ b/docs/userguide/install.md
@@ -16,20 +16,26 @@ It is recommended to install utPLSQL in it's own schema. You are free to choose 
 
 The installation user/schema must have the following Oracle system permissions during the installation.
   - CREATE SESSION
+  - CREATE SEQUENCE
   - CREATE PROCEDURE
   - CREATE TYPE
   - CREATE TABLE
   - CREATE VIEW
   - CREATE SYNONYM
+  - SELECT ANY DICTIONARY
   - ALTER SESSION
   
 In addition it must be granted execute to the following system packages.
 
   - DBMS_LOCK
+  - DBMS_LOB
+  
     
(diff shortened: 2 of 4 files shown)


.travis/install.sh
```diff
@@ -9,8 +9,6 @@ set feedback off
 set verify off
 
 @../source/create_utplsql_owner.sql $UT3_OWNER $UT3_OWNER_PASSWORD $UT3_TABLESPACE
---only needed to run unit tests for utplsql v3, not required to run utplsql v3 itself
-grant select any dictionary to $UT3_OWNER;
```
Pazus: lets try granting `SELECT_CATALOG_ROLE` role instead. I think it'll be enough

docs/userguide/install.md
```diff
@@ -16,20 +16,26 @@ It is recommended to install utPLSQL in it's own schema. You are free to choose
 
 The installation user/schema must have the following Oracle system permissions during the installation.
   - CREATE SESSION
+  - CREATE SEQUENCE
   - CREATE PROCEDURE
   - CREATE TYPE
   - CREATE TABLE
   - CREATE VIEW
   - CREATE SYNONYM
+  - SELECT ANY DICTIONARY
```
Pazus: Its too strong... i think `SELECT_CATALOG_ROLE` is enough

source/create_utplsql_owner.sql
```diff
@@ -27,9 +27,10 @@ define ut3_tablespace = &3
 
 create user &ut3_user identified by &ut3_password default tablespace &ut3_tablespace quota unlimited on &ut3_tablespace;
 
-grant create session, create sequence, create procedure, create type, create table, create view, create synonym to &ut3_user;
+grant create session, create sequence, create procedure, create type, create table, create view, create synonym, select any dictionary to &ut3_user;
```
Pazus: As this is only used by coverage reporter I dont think we should have it here. There will be lots of DBAs who wount give this grant even its on test server.

jgebal: Aren't we're trying to satisfy DBA's needs too much?
I fear that If we will try to satisfy everyone's needs we will not progress and keep on stumbling upon issues.
We would probably need to separate coverage reporting capability from the main framework and make it a separate add-on but I don't really like the idea.

**How should we separate this privilege in your opinion?**

We could change the privilege to be one of:
- `select catalog role`
- `select on dba_source` (preffered)
- `select on all_source` (not so good, but used currently)
- `select_any_dictionary` (not tested)

Shoelace: or not grant the privilege at all.. and use `AUTHID CURRENT_USER` and required that the user running the test has the correct permissions .

jgebal:
I've found a different solution, so moving away from this one.

Issue #267, opened 2017-04-06
Suite manager cache is not invalidated after package is dropped


jgebal:
If a schema was scanned, the suite manager keeps the cached suite for that schema.
If a ut package is dropped, the manager does not see it as a change and is not scanning the schema for suites again.

Suite manager cache needs to be invalidated in following cases:
- package added - currently supported
- package modified - currently supported
- package removed - **currently not supported**

PR #274, opened 2017-04-12
fix cache invalidation on drop

Pazus:
Fix cache invalidation on package drop
Addresses #267


jgebal:
Looks like suite manager doesn't line the $ sign inside the test package name.
I'll fix the test and approve.
I'm not sure if it is expected / good practice to put $ sign into test package names therefore will not treat it as an error for now.

jgebal:
Resolves #267

PR #295, opened 2017-04-29
Adding source_path and test_path parameters for coverage reporters

### LINKED ISSUE #212: Fixes to ut_run.sql
- [ ] Update of RunAll.sql to dynamically generate project files list.
- [ ] Add support for coverage parameters.

### LINKED ISSUE #241: Add support for project sources path to ut_run.sql for coverge reporting
The `ut_run.sql` should accept additional parameter to indicate project sources path.
The path parameter should be used with command line utility (dir-windows/ls-linux) to list all files in that path.
The files then should become part of the invocation of coverage reporter(s).

viniciusam:
Needs testing on linux environment.
Resolves #241 
Resolves #212


jgebal:
This looks really great, thanks @viniciusam 
Can you add some error handling into `ut_run.bat` and `ut_run.sh`?
Can you update documentation for [ut_run-script](https://github.com/utPLSQL/utPLSQL/blob/develop/docs/userguide/ut_run-script.md)
That would be greatly appreciated as it would make this change complete.
It would be really handy if you could document the basic steps around executing the new ut_run scripts.
From what I know, they need to be executed from the project root path, right?

viniciusam:
> Can you update documentation for ut_run-script

ok!

> From what I know, they need to be executed from the project root path, right?

Yes, if you are running with coverage reporters and need to pass source_path and/or test_path. Otherwise, you can run it from anywhere.

client_source/sqlplus/ut_run.sql
```diff
@@ -70,7 +85,12 @@ Parameters:
 
 whenever sqlerror exit failure
 whenever oserror exit failure
-conn &1
+
+define client_path="&1"
```
Pazus: accordint to the invokation example in the beginning of the file (`ut_run.sql user/password@database [-p=(ut_path|ut_paths)] [-c] [-f=format [-o=output] [-s] ...]` 
&1 corresponds to user/password@database and all the others are wrong...

client_source/sqlplus/ut_run.sql
```diff
@@ -167,17 +189,14 @@ declare
...
-      with
-        param_vals as(
-          select regexp_substr(column_value,'-([fos])\=?(.*)',1,1,'c',1) param_type,
-                 regexp_substr(column_value,'-([fos])\=(.*)',1,1,'c',2) param_value
-          from table(a_params)
-          where column_value is not null)
-      select param_type, param_value
-      from param_vals
-      where param_type is not null
+      select regexp_substr(column_value,l_param_regex,1,1,'c',1) param_type,
```
Pazus: change is not equivalient `regexp_substr` will return null if string doesn't match the pattern

client_source/sqlplus/ut_run.sql
```diff
@@ -289,7 +355,25 @@ begin
...
-        p('  v_reporter := '||l_reporter_name||'();');
+        if lower(l_reporter_name) in ('ut_coveralls_reporter', 'ut_coverage_sonar_reporter', 'ut_coverage_html_reporter') then
+          p('  v_reporter := '||l_reporter_name||'( a_file_paths => ut_varchar2_list(');
+          loop
+            fetch :l_source_files into l_file_path;
+            exit when :l_source_files%notfound or l_file_path is null;
+            p('      '''||l_file_path||''',');
+          end loop;
+            p('      null));');
+        elsif lower(l_reporter_name) = ('ut_sonar_test_reporter') then
```
Pazus: is it the same as the `IF`?

client_source/sqlplus/ut_run.sql
viniciusam: yes, but I changed the where condition to bring only ones that match the pattern:
```
select regexp_substr(column_value,l_param_regex,1,1,'c',1) param_type,
             regexp_substr(column_value,l_param_regex,1,1,'c',3) param_value
      from table(a_params)
      where column_value is not null
        and regexp_like(column_value,l_param_regex)
```

client_source/sqlplus/ut_run.sql
viniciusam: Will update that, thanks!

client_source/sqlplus/ut_run.sql
viniciusam: Didn't get it...

client_source/sqlplus/ut_run.sql
viniciusam: We have the same documentation on `ut_run-script.md`, maybe we should remove this and let a link to the one in the userguide. What do you think?

jgebal:
I got following error when testing with SQLcl on Ubuntu:
```
jacek@jacek-VirtualBox:~/utPLSQL/utPLSQL/client_source/sqlplus$ bash ./ut_run ut3/ut3@localhost:1521/orclpdb1

SQLcl: Release 4.2.0 Production on Fri May 05 12:51:18 2017

Copyright (c) 1982, 2017, Oracle.  All rights reserved.

Connected.
bash: /home/jacek/utPLSQL/utPLSQL/client_source/sqlplus\file_list.bat: No such file or directory
```
You need to use `/` instead of `\` for Unix.

jgebal:
Actually the syntax is OK, but you need to use:
`file_list.bat ... | echo ""` so that it does not fail if it can't execute.
After doing that change i now get:
```
jacek@jacek-VirtualBox:~/utPLSQL/utPLSQL/client_source/sqlplus$ bash ./ut_run ut3/ut3@localhost:1521/orclpdb1

SQLcl: Release 4.2.0 Production on Fri May 05 12:57:00 2017

Copyright (c) 1982, 2017, Oracle.  All rights reserved.

Connected.
bash: /home/jacek/utPLSQL/utPLSQL/client_source/sqlplus\file_list.bat: No such file or directory


bash: /home/jacek/utPLSQL/utPLSQL/client_source/sqlplus\file_list.bat: No such file or directory
bash: /home/jacek/utPLSQL/utPLSQL/client_source/sqlplus/file_list: Permission denied

bash: /home/jacek/utPLSQL/utPLSQL/client_source/sqlplus/file_list: Permission denied


```

viniciusam:
What happens if you add the client to the env, and execute ut_run like this?
```
export PATH=~/utPLSQL/utPLSQL/client_source/sqlplus:$PATH
cd ~
ut_run ut3/ut3@localhost:1521/orclpdb1
```

client_source/sqlplus/ut_run.sql
jgebal: I like the idea of keeping documentation in one place but it is of great value to have full syntax visible from within script.
It would be probably best to have documentation in both places, even if it means double maintenance.
Ideally we should have `ut_run.bat` and `ut_run.sh` report help on syntax if the syntax provided is not valid. Documentation in readme might be redundant then, as the scripts would be self-documenting.
Any thoughts on that?

client_source/sqlplus/ut_run.sql
jgebal: The previous regex was not working fine when using `-source_path=...` param.
The param `-source_path=` was enabling the `-s` flag too, which was clearly a bug in the param parsing.

client_source/sqlplus/ut_run.sql
jgebal: It is not, as `ut_sonar_test_reporter` accepts list of file names that represent test code. The ones in `if` are passing file names for project sources (code under test).

docs/userguide/ut_run-script.md
```diff
@@ -1,52 +1,71 @@
...
+                      -f=ut_coverage_html_reporter
+                        Generates a HTML coverage report providing summary    and detailed information on code coverage.
+                        The html reporter is based on open-source    simplecov-html reporter for Ruby.
+                        It includes source code of the code that was covered   (if possible).
+                      -f=ut_coveralls_reporter
+                        Generates a JSON coverage report providing detailed    information on code coverage with line numbers.
+                        This coverage report is designed to be consumed by    cloud services like https://coveralls.io/.
+                      -f=ut_coverage_sonar_reporter
+                        Generates a JSON coverage report providing detailed    information on code coverage with line numbers.
+                        This coverage report is designed to be consumed by    local services like https://about.sonarqube.com/.
```
jgebal: I would replace the line
This coverage report is designed to be consumed by    local services like https://about.sonarqube.com/.
with something like:
This report is designed to be consumed by [SonarQube](https://about.sonarqube.com/) coverage.

docs/userguide/ut_run-script.md
```diff
@@ -1,52 +1,71 @@
...
+                      -f=ut_sonar_test_reporter
+                        Generates a JSON report providing detailed     information on test specifications.
+                        This report is designed to be consumed by local     services like https://about.sonarqube.com/.
```
jgebal: I would replace the line
This report is designed to be consumed by    local services like https://about.sonarqube.com/.
with something like:
This report is designed to be consumed by [SonarQube](https://about.sonarqube.com/) to report on test execution.

client_source/sqlplus/ut_run.bat
```diff
@@ -0,0 +1,16 @@
...
+    echo Error: ut_run script path cannot have spaces.
+    exit /b 1
+)
+
+if "%1" == "" (
+    echo Usage: ut_run user/password@database [options...]
+    exit /b 1
+)
+
+sqlplus /nolog @"%clientDir%\ut_run.sql" '%clientDir%' '%projectDir%' %*
```
jgebal: since we now wrap the sqlplus call (user does not call sqlplus but call bash/cmd script instead), could we make the scripts in a way that they can run using both sqlplus and sqlcl?
could `sqlplus` / `sqlcl` be parameters?

client_source/sqlplus/ut_run.sql
viniciusam: We could remove only the invocation examples from ut_run.sql, as they are already in readme.

client_source/sqlplus/ut_run.bat
viniciusam: I need to think more about how to implement that.

viniciusam:
Do you have any thoughts on how to use the file_list script on tests/RunAll.sql?

I am thinking of making the clientDir of ut_run.bat/sh an evironment var.
That way we won't need to pass it as a parameter to file_list and file_list can check if it is not empty to use it. If it is empty (called directly or from a different script), it will use it's own path.
The way it is now, the tests/RunAll.sql will need to pass the clientDir to use the file_list script.

jgebal:
We should follow the concept of configuration by exception:
http://stackoverflow.com/questions/31406903/configuration-by-exception-vs-convention-over-configuration
So I would go with following:
If no paths defined assume:
- tests are in test directory 
- sources are in source directory 
User should be able to use both eng variables and parameters.
The priority would be: parameters override environment variables, environment variables override defaults.

viniciusam:
- Removed the client_path parameter from file_list script, now the user can pass the full output file path. If just a filename is passed, the file will be created on caller's dir.
- Assume source and tests as default.

PR #300, opened 2017-05-01
Changed release archive to be version-agnostic, so that it is possibl…

jgebal:
…e to download and install the most recent version using `github/latest release link

### DIFF
--- a/.travis/build_release_archive.sh
+++ b/.travis/build_release_archive.sh
@@ -4,7 +4,7 @@ mv -f .gitattributes.release .gitattributes
 git add .
 git commit -m "tmp commit for building a release archive"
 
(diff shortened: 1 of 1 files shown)


Pazus:
What is the name of the archive now?

jgebal:
It's called `utPLSQL` now

Pazus:
Could you show an URL how it will look like?
BTW:
https://github.com/facebook/react/releases

jgebal:
I've found a way to download latest release without need to have the release file version-agnostic.
```bash
file_type='zip' # either 'zip' or 'gz' is availiable
download_url=`curl --silent https://api.github.com/repos/utPLSQL/utPLSQL/releases/latest | /usr/bin/awk '/browser_download_url/ { print $2 }' | grep ".${file_type}" | /usr/bin/sed 's/"//g'`
# echo ${download_url} # for debugging purposes
curl -LOk ${download_url}
```

jgebal:
Since we don't need to make files version-agnostic to make them downloadable, I'm closing this PR.

PR #302, opened 2017-05-06
Refactored reporters API so that you can use API in unified way regar…

### LINKED ISSUE #272: narrow down constructors for sonar/coveralls coverage reporters
coverage reporters used for sonar and coveralls only bring value if they are reporting on project files.
The constructors that do not accept files list do not make much sense and can be confusing for users/integrators.
It's best to remove them while we're still at beta.

### LINKED ISSUE #279: Improve file mapping API to make it easier to use by consumers
We should redesigning API for reporters. Current solution, were we are passing parameters to reporter constructor (file mapping) makes API usage complex. Each reporter can have different constructor and this makes API hard to integrate.

I'm thinking of two options:
1. Make a set of abstract procedures `init` to initialize the file mappings/file lists in the `ut_reporter_base` class. This way that every reporter would need to implement it and the API for invoking reporters would become consistent by rule. The downside of this approach however is that many reporters would need to implement the `init` procedures with NULL implementation if the procedure is not relevant. Also, many reporters would end up having duplicated implementation for the `init` procedures (as it is currently with constructors).
2. Extend the `ut_runner.run` and `ut.run` to accept the file mappings/file lists. This sounds better solution as the parameters would become part of `ut_run` object and could be consumed by any reporter according to it's needs.

I lean towards solution number 2.
The `run` procedures in `ut` and `ut_runner` packages would be extended by 2 optional parameters: `a_source_file_mappings, a_test_file_mappings`
The mapper would be made available to API consumers so that we would still have three different ways to build file mappings:
1. by providing full mapping of file to db object
2. by building mapping using just file names and relying on default mapping rules
3. by building mapping using custom regex rules to parse filename/path and determine database object from it

We should perform the refactoring as soon as possible so we minimize the impact on the work getting cone on the utPLSQL-java-aip project.

jgebal:
…dless of reporter that is used.

All reporters are now parameter-agnostic and the parameters are defined at the `ut_run` level.
Exception is ut_coverage_html_reporter, which is using two individual parameters (a_html_report_assets_path, a_project_name), but those are optional.

Resolves #279
Resolves #272


jgebal:
Please review.
A documentation update will be needed for this.
Also some additional tests for `ut.run` API end points would be nice to have.

source/core/coverage/ut_coverage.pkb
```diff
@@ -275,23 +165,22 @@ create or replace package body ut_coverage is
...
     l_query            varchar2(32767);
   begin
+    l_schema_names := coalesce(a_coverage_options.schema_names, ut_varchar2_list(sys_context('USERENV','CURRENT_SCHEMA')));
 
     if not ut_coverage_helper.is_develop_mode() then
       --skip all the utplsql framework objects and all the unit test packages that could potentially be reported by coverage.
-      l_skipped_objects := ut_utils.get_utplsql_objects_list()
-                           multiset union all g_unit_test_packages
-                           multiset union all coalesce(g_exclude_list, ut_object_names());
+      l_skipped_objects := ut_utils.get_utplsql_objects_list() multiset union all coalesce(a_coverage_options.exclude_objects, ut_object_names());
```
Pazus: It was `g_unit_test_packages` here previously, why is it removed from the skipped item list?

source/core/ut_file_mapper.pkb
```diff
@@ -0,0 +1,95 @@
+create or replace package body ut_file_mapper is
```
Pazus: like the idea of responsibility spread

PR #315, opened 2017-05-17
Fix typo

paytonrules:
I haven't checked the license, but I doubt I'm allowed to sue the framework.


jgebal:
Thanks, good catch.

PR #325, opened 2017-05-19
Fix HTML Coverage

lwasylow:
Fix to html coverage to allow overwrite default schema on test path to get coverage from user supplied schemas as well.


lwasylow:
Merge into wrong branch

PR #326, opened 2017-05-19
Fix to HTML Coverage

lwasylow:
Adding ability to overwrite default schemas list used in html code coverage.


source/core/types/ut_run.tpb
```diff
@@ -29,7 +29,8 @@ create or replace type body ut_run as
     self.coverage_options := a_coverage_options;
     self.test_file_mappings := coalesce(a_test_file_mappings, ut_file_mappings());
     if self.coverage_options is not null then
-      l_run_schemes := get_run_schemes();
+	  l_run_schemes := get_run_schemes();
+      l_run_schemes := coalesce(coverage_options.schema_names, l_run_schemes);
```
jgebal: It seems that this fix will work well but the variable `l_run_schemes` changes it's meaning.
Before that it was holding list of schema names that are holding unit tests to run.
With this operation, it will hold the schema names that contain code to gather coverage on.
It actually seems a right thing to do but I would change the variable name.
It should be called `l_coverage_schema_names` or `l_schema_names_to_cover` or something like this.
This way we avoid confusion and ambiguity.

Could you also make those two lines into one with single assignment and make sure indents are consistent?

PR #338, opened 2017-06-01
Download URL Variables

lovromazgon:
A small fix in the documentation, changing the name of the download URL variables.


jgebal:
Thank you @lovromazgon

PR #344, opened 2017-06-07
Improved failure message when comparing incomparable data-types.

### LINKED ISSUE #334: Bad/misleading failure message when comparing different datatypes
Currently when comparing two different datatypes for equality we get something like below:
```
5) object_compare_different_type
      The matcher 'equal' cannot be used with data type (ut3.demo_department_new).
      at "UT3.DEMO_EQUAL_MATCHER", line 69 ut.expect(anydata.convertObject(l_actual)).to_equal(anydata.convertObject(l_expected));
```
You can see it on [travis-ci build](https://travis-ci.org/utPLSQL/utPLSQL/jobs/237259472#L1579)

This should be more of a message like:
```
      The matcher 'equal' cannot be used to compare datatypes (ut3.demo_department_new) and (ut3.demo_department).
```

The message is misleading and could make developer think that the matcher is not capable of evaluating the given datatype.

jgebal:
Fixes #334


jgebal:
@Pazus, are you ok with this change?

PR #349, opened 2017-06-11
Minor updates to the MD documentation files

john-otoole:
I've made a bunch of minor updates to the markdown files based on observations from reading them for the first time.
Most of them are trivial updates, but check annotations.md in particular as was unsure about the third bullet point of section "A Suitepath can be provided in three ways".


jgebal:
Thanks. It is great to have someone native review the doc.

john-otoole:
No problem, I'll update anything else I spot as I start using utPLSQL.  It
looks like a comprehensive framework - very impressive.

John

PR #362, opened 2017-06-23
Install and Uninstall scripts are now much more readable

### LINKED ISSUE #321: Improve feedback readability from install scripts
The install scripts should provide nice feedback, as it was in v2.

jgebal:
Resolves #321


jgebal:
Hi @Pazus, to get this done, we would need to get rid of the `show errors`.
That could be fine, as we still display the summary of all errors at the end.

We might want to tweak this so that the error report is more readable.
One more thing to change would be to conditionally deply the `proftab.sql`
It should only be installed only if one of profiler tables or the sequence is missing.
This way we will have install script running without any errors.

jgebal:
@Pazus , your comment should now be adressed.
Please review.

Issue #366, opened 2017-06-26
Character string buffer too small (ut_utils.clob_to_table)


dmvolodin:
For packages containing multibyte characters we get that (ut_coverage_html_reporter)
```
ORA-06502: PL/SQL: numeric or value error: character string buffer too small
ORA-06512: at "UT3.UT_RUNNER", line 87
ORA-06512: at "UT3.UT_RUNNER", line 111
ORA-06512: at "UT3.UT", line 292
ORA-06512: at "UT3.UT", line 267
ORA-06512: at line 2
```
Oracle 11.2 docs:

> **Declaring Variables for Multibyte Characters**
> The maximum size of a CHAR or VARCHAR2 variable is 32,767 bytes, whether you specify the maximum size in characters or bytes.
> When declaring a CHAR or VARCHAR2 variable, to ensure that it can always hold n characters in any multibyte character set, declare its length in characters—that is, CHAR(n CHAR) or VARCHAR2(n CHAR), where n does not exceed FLOOR(32767/4) = 8191.

Declaring with 32767
`function clob_to_table(a_clob clob, a_max_amount integer := 32767, a_delimiter varchar2:= chr(10)) return ut_varchar2_list`

Simple testcase
1. Single-byte "q"
```
declare
  c varchar2(32767 char);
begin
  c := rpad('q',32767,'q');
end;
/
```
OK
2. Multibyte "Ж" (Cyrillic)
```
declare
  c varchar2(32767 char);
begin
  c := rpad('Ж',32767,'Ж');
end;
/ 
```
Failed (numeric or value error: character string buffer too small)
3. Multibyte "Ж" (Cyrillic), rpad for 32k/2
```
declare
  c varchar2(32767 char);
begin
  c := rpad('Ж',32767/2-1,'Ж');
end;
/  
```
OK

jgebal:
@dmvolodin 
Can you provide full stack trace, that is printed to DBMS_OUTPUT?
The stack trace provided is the outer trace (from WHEN OTHERS).

I'm aware of the CHAR/BYTE issue with VARCHAR2 length. I'm just curious what was the scenario that caused this issue to occur.

Would you be able to provide a reproducible scenario, so that we can build a test case out of it?
Would you be willing to create a test case and a fix?

It is quite inconvenient to limit the length to 1/4th of max size for all deployments and all databases. At the moment however I don't know if there is any better option, so that we limit the size according to the content type (single/multi-byte).

[...]

PR #369, opened 2017-06-27
dbms_lob_read_multibyte/Fix multibyte issue with dbms_lob.read

dmvolodin:
Use (3/4) * 32767 instead of 32767 as buffer default size for dbms_lob.read


dmvolodin:
Fix for #366

dmvolodin:
OK, I'll change it

PR #371, opened 2017-06-27
Enriched warning message on implicit/explicit commit

### LINKED ISSUE #364: Improve message on failure of rollback(auto)
Currently the message says:
`Savepoint not established. Implicit commit might have occured.`
That message is not saying what is wrong to the users.
A message like:
```
Unable to perform automatic rollback after procedure: my_demo_test_package.cleanup
An implicit or explicit commit/rollback occurred.
Use the %rollback(manual) annotation or remove commits/rollback/ddl statements that are causing the issue.
```

To be able to point to a point, where rollback was to be performed, we need to pass the `ut_executable` to `rollback_to_savepoint` procedure.

Pazus:
Resolves #364


jgebal:
I was thinking of following approach:
In ut_suite_item have `start_transaction_id`. 
The value of that variable would be set when savepoint is created (using `DBMS_TRANSACTION.LOCAL_TRANSACTION_ID`)
Inside `ut_executable` we can reference the value of `start_transaction_id` from the `self` passed by invoking test/suite.
We could then check in the `ut_executeable`, after executing block, if transaction id is still the same.
If it's NULL or it is changed, commit/rollback was done.

This way we can identify exactly where the savepoint was broken and provide a proper error/failure/warning message.

Pazus:
Started implementation of the requested changes and realized its tricky

if a test has commit, I can find the place of the commit in `ut_executables` of the test. but then if we has a stacked suites rollback after `afterall` of the top-level suite will also fail and I dont see a way to propagate the place of commit there...

Do you think we need it? Also several commits might occure, do we need to list them all?

jgebal:
The executable has a flag indicating that commit occurred.
The test/suite checks if it is rollback auto.
For each executable if rollback auto and commit occurred, print warning.
We should then be able to remove warning from the rollback procedure itself.
I think we should print all warnings so that user can fix it all in one shot

jgebal:
**Current proposition for warnings implementation.**

For the code:
```sql
create or replace package ut_output_test_rollback as
  --%suite

  --%beforeall
  procedure ba;
  --%beforeeach
  procedure be;
  --%test
  procedure test;
  --%test
  procedure test1;
  --%test
  --%rollback(manual)
  procedure t_manual;
  --%afterall
  procedure aa;
  --%aftereach
  procedure ae;
end;
/

create or replace package body ut_output_test_rollback as
 procedure ba is begin commit; end;
 procedure be is begin commit; end;
 procedure test is begin commit; end;
 procedure test1 is begin commit; end;
 procedure t_manual is begin commit; end;
 procedure aa is begin commit; end;
 procedure ae is begin commit; end;
end;
/
```

You get outputs from: `select * from table(ut.run('ut_output_test_rollback'));`

```
ut_output_test_rollback
  test [.007 sec]
  test1 [.006 sec]
  t_manual [.005 sec]
 
 
Warnings:
 
  1) ut_output_test_rollback.test
      Unable to perform automatic rollback after test. An implicit or explicit commit/rollback occurred in procedures:
        ut3.ut_output_test_rollback.be
        ut3.ut_output_test_rollback.test
        ut3.ut_output_test_rollback.ae
      Use the "--%rollback(manual)" annotation or remove commit/rollback/ddl statements that are causing the issue.
 
  2) ut_output_test_rollback.test1
      Unable to perform automatic rollback after test. An implicit or explicit commit/rollback occurred in procedures:
        ut3.ut_output_test_rollback.be
        ut3.ut_output_test_rollback.test1
        ut3.ut_output_test_rollback.ae
      Use the "--%rollback(manual)" annotation or remove commit/rollback/ddl statements that are causing the issue.
 
  3) ut_output_test_rollback
      Unable to perform automatic rollback after test suite. An implicit or explicit commit/rollback occurred in procedures:
        ut3.ut_output_test_rollback.ba
        ut3.ut_output_test_rollback.aa
        ut3.ut_output_test_rollback.be
        ut3.ut_output_test_rollback.test
        ut3.ut_output_test_rollback.ae
        ut3.ut_output_test_rollback.test1
        ut3.ut_output_test_rollback.t_manual
      Use the "--%rollback(manual)" annotation or remove commit/rollback/ddl statements that are causing the issue.
 
Finished in .028771 seconds
"3 tests, 0 failed, 0 errored, 0 disabled, 3 warning(s)"
```

source/core/types/ut_executable.tpb
```diff
@@ -135,9 +136,14 @@ create or replace type body ut_executable is
 
       l_completed_without_errors := (self.error_stack||self.error_backtrace) is null;
 
-      a_listener.fire_after_event(self.associated_event_name, a_item);
       --listener - after call to executable
+      a_listener.fire_after_event(self.associated_event_name, a_item);
+
+      if l_start_transaction_id != dbms_transaction.local_transaction_id(true) then
```
jgebal: If `beforetest` commits and `test` commits, we cannot "see" that the test committed as there was no transaction (it was committed at the time we checked).

jgebal: If `beforetest` commits and `test` commits, we cannot "see" that the test committed as there was no transaction (it was committed at the time we checked).

source/core/ut_annotations.pkb
```diff
@@ -78,7 +78,7 @@ create or replace package body ut_annotations as
                                                       ,'%(' || c_rgexp_identifier || ')'
                                                       ,modifier => 'i'
                                                       ,subexpression => 1));
-        l_annotation_params_str := trim(regexp_substr(l_annotation_str, '\((.*?)\)$', subexpression => 1));
+        l_annotation_params_str := trim(regexp_substr(l_annotation_str, '\((.*?)\)\s*$', subexpression => 1));
```
jgebal: Could go as a separate PR for traceability

jgebal:
@Pazus 
Can you review the existing solution and let me know if it's OK with you?
I think it's fine to spit out all the warnings, even if there is plenty of them.
My reasoning is that developers should fix them and avoid using `--%rollback(auto)`  with transaction controlled tests/code.

source/core/types/ut_executable.tpb
Pazus: But then function will return null, we should check for null.

`if l_start_transaction_id != dbms_transaction.local_transaction_id Or dbms_transaction.local_transaction_id is null`

jgebal: I mean, we need to create transaction at the start.
You're right, we don't need to create it at the end, but do an additional null check.
If you think it should be changed, so that we don't create transaction after execution, can you update the PR?
I'm ok with both solutions.

Issue #351, opened 2017-06-14
Add timing on individual tests to documentation reporter


jgebal:
It would be nice to be able to see timing on individual tests executed by documentation reporter.
That could be done as:
- separate reporter on top of the `ut_documentation_repoter`
- modification of the `ut_documentation_repoter`

PR #372, opened 2017-06-27
implemented timing reporting for documentation reporter

Pazus:
Added timing reporting for documentation reporter (nonconditional).
addressed #351


jgebal:
I wonder if reporter with timing should be a separate reporter (under `ut_documentation_reporter`).
The new reporter would only need to override the `after_calling_test` method.
Also, maybe use square brackets to make `[.012 sec]` more distinguishable from `(FAILURE)`

Pazus:
Do you think we need a separate one? At first I thought the same but then I realized that we go that we then we have to make a choice which reporter to use, and as its mostly used from `ut.run` we don't have a way to choose there. This means standard `ut.run` reporter has to be the most appropriate for the human-user (not CI). I think timing is a simple nice addition.

Like your idea about square brackets! Changing it.

Issue #350, opened 2017-06-14
documentation reporter warnings


jgebal:
There is a little bug in documentation reporter.
The warnings are not numbered properly, and every warning has number: 1
Example:
```
Warnings:
 
  1) my_suite.ut_package1
      Savepoint not established. Implicit commit might have occured.
 
  1) my_suite.ut_package2
      Savepoint not established. Implicit commit might have occured.
 
  1) my_suite.ut_package3
      Savepoint not established. Implicit commit might have occured.
 
...
```

PR #373, opened 2017-06-27
Fixed indexing of warnings for documentation reporter

Pazus:
address #350


source/reporters/ut_documentation_reporter.tpb
```diff
@@ -151,7 +153,8 @@ create or replace type body ut_documentation_reporter is
 
       if a_item.warnings is not null and a_item.warnings.count > 0 then
         for i in 1 .. a_item.warnings.count loop
-          self.print_text('  ' || i || ') ' || a_item.path);
+          l_warning_index := l_warning_index + 1;
+          self.print_text('  ' || l_warning_index || ') ' || a_item.path);
```
jgebal: Can you make both `error_no` and `warning_no` uniformly indented
```
lpad(l_warning_index,5)
```
I don't expect any report to have more than 9999 errors or warnings in it.

Pazus: Done. It is already done for errors, so I made it only for warnings

Issue #357, opened 2017-06-16
Update Documentation - Annotation Section


lwasylow:
Annotation -- %suitepath cannot have a whitespace after it.

Please update documantation.

Also there seems to be incosistency as the whitespace is allowed on annotation of test etc.

Pazus:
Hi, @lwasylow 

Could you provide and example of a broken annotation?

lwasylow:
Hi Pazus
Generally I think issue is that approach is not consistent.When using a suitepath annotation it causes an error that tells about null index rather than highlight syntax issues.
On the other hand a displayname annotation when provided with whitespace results in no error just dont display a test name on result.

E.g.
%suitepath (alltests.test_setup)

In my opinion I think the approach should be a consistent so regexp should be modified to throw error when whitespace is found with Syntax error info.

PR #375, opened 2017-06-28
Fix error with space before annotation params

Pazus:
addresses #357

### DIFF
--- a/source/core/ut_annotations.pkb
+++ b/source/core/ut_annotations.pkb
@@ -30,7 +30,7 @@ create or replace package body ut_annotations as
   c_rgexp_identifier            constant varchar2(50) := '[a-z][a-z0-9#_$]*';
   c_annotation_block_pattern    constant varchar2(200) := '(({COMMENT#.+}'||chr(10)||')+)( |'||chr(09)||')*(procedure|function)\s+(' ||
                                                            c_rgexp_identifier || ')';
-  c_annotation_pattern          constant varchar2(50) := gc_annotation_qualifier || c_rgexp_identifier || '(\(.*?\)$)?';
+  c_annotation_pattern          constant varchar2(50) := gc_annotation_qualifier || c_rgexp_identifier || '[ '||chr(9)||']*(\(.*?\)$)?';
 
 
   function delete_multiline_comments(a_source in clob) return clob is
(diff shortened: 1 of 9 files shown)


tests/ut_suite_manager/ut_suite_manager.emptySuitePath.sql
```diff
@@ -0,0 +1,33 @@
...
+set termout on
+
+declare
+  l_objects_to_run ut_suite_items;
+begin
+
+  --act
+  l_objects_to_run := ut_suite_manager.configure_execution_by_path(ut_varchar2_list('tst_empty_suite_path'));
+
+  if ut_expectation_processor.get_status = ut_utils.tr_success then
```
jgebal: I don't understand how will this test ever fail, as there is no expectation executed before status is checked.

tests/ut_annotations/ut_annotations.parse_package_annotations.spaceBeforeAnnotationParams.sql
```diff
@@ -0,0 +1,39 @@
...
+
+  procedure foo;
+END;';
+
+--Act
+  l_parsing_result := ut_annotations.parse_package_annotations(l_source);
+
+--Assert
+  l_ann_param := null;
+  l_ann_param.val := 'Name of suite';
```
jgebal: Lines 24 and 25 are not used I think

source/core/ut_annotations.pkb
```diff
@@ -30,7 +30,7 @@ create or replace package body ut_annotations as
   c_rgexp_identifier            constant varchar2(50) := '[a-z][a-z0-9#_$]*';
   c_annotation_block_pattern    constant varchar2(200) := '(({COMMENT#.+}'||chr(10)||')+)( |'||chr(09)||')*(procedure|function)\s+(' ||
                                                            c_rgexp_identifier || ')';
-  c_annotation_pattern          constant varchar2(50) := gc_annotation_qualifier || c_rgexp_identifier || '(\(.*?\)$)?';
+  c_annotation_pattern          constant varchar2(50) := gc_annotation_qualifier || c_rgexp_identifier || '[ '||chr(9)||']*(\(.*?\)$)?';
```
jgebal: Shouldn't we use `\s*` rather than `chr(9)`?
We should also enhance the pattern to allow \s at the end I think.
So how about:
```sql
gc_annotation_qualifier || c_rgexp_identifier || '\s*(\(.*?\)\s*$)?';
```

Pazus: `\s` contains new line character which is not allowed

jgebal: Ok

jgebal: Can you make the regexp work with trailing spaces?
'[ '||chr(9)||]*(\(.*?\)[ '||chr(9)||']*$)?'

PR #377, opened 2017-06-29
Use new image for 12c-r1

viniciusam:
to solve disk space issues


jgebal:
While you're on it.
Do you have any ideas to improve performance of the 12 databases?
They run very slow.

viniciusam:
Don't think we have much to do on Travis, maybe try a container native tool.
I was testing wercker and circleci v2, seems to be good alternatives.

I also think we should have those images dockerfiles on our repository, I can work on that.

Issue #383, opened 2017-07-03
Use install_headless to install on Travis


Pazus:
Use install_headless script to install on Travis

Issue #382, opened 2017-07-03
Dont show error when profiler tables exist


Pazus:
Fix the installation script not to show ugly errors when profiler tables already exist in the schema.

PR #384, opened 2017-07-03
Improved installation routine

Pazus:
addressed #383
addressed #382

### DIFF
--- a/source/core/coverage/proftab.sql
+++ b/source/core/coverage/proftab.sql
@@ -33,7 +33,7 @@ begin
   select count(*) into l_tab_exist from 
   (select table_name from all_tables where table_name = 'PLSQL_PROFILER_UNITS' and owner = sys_context('USERENV','CURRENT_SCHEMA')
    union all
-   select synonym_name from user_synonyms where synonym_name = 'PLSQL_PROFILER_UNITS' and owner = sys_context('USERENV','CURRENT_SCHEMA'));
+   select synonym_name from all_synonyms where synonym_name = 'PLSQL_PROFILER_UNITS' and owner = sys_context('USERENV','CURRENT_SCHEMA'));
   if l_tab_exist = 0 then
     execute immediate q'[create table plsql_profiler_units
 (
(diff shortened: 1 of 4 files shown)


source/core/coverage/proftab.sql
```diff
@@ -1,4 +1,9 @@
-create table plsql_profiler_runs
+declare
```
jgebal: I'm still thinking if we should make scripts rerunable or drop objects on uninstall.
Also, since we modify script provided from Oracle,  should we split it into one script per object?

Pazus: I think we shouldn't drop it. Could you explain why do you want to split it?

jgebal: We use a pattern of one script for one object so this script is the only one breaking the pattern

Pazus: but its an oracle object, not our... But I can fix it tonight

jgebal: Ok, lets leave it as it is

source/core/coverage/proftab.sql
```diff
@@ -1,4 +1,12 @@
...
+  l_tab_exist number;
+begin
+  select count(*) into l_tab_exist from 
+  (select table_name from all_tables where table_name = 'PLSQL_PROFILER_RUNS' and owner in (user,'PUBLIC')
```
jgebal: We cannot have any table with owner = 'PUBLIC'

source/core/coverage/proftab.sql
```diff
@@ -31,12 +50,23 @@ create table plsql_profiler_units
...
+  end if;
+end;
+/
 
-create table plsql_profiler_data
+declare
+  l_tab_exist number;
+begin
+  select count(*) into l_tab_exist from 
+  (select table_name from all_tables where table_name = 'PLSQL_PROFILER_DATA' and owner in (user,'PUBLIC');
```
jgebal: semicolon at the end.

source/core/coverage/proftab.sql
```diff
@@ -52,10 +82,25 @@ create table plsql_profiler_data
...
+  end if;
+end;
+/
 
-create sequence plsql_profiler_runnumber start with 1 nocache;
+declare
+  l_seq_exist number;
+begin
+  select count(*) into l_seq_exist from 
+  (select sequence_name from all_sequences where sequence_name = 'PLSQL_PROFILER_RUNNUMBER' and sequence_owner in (user,'PUBLIC');
```
jgebal: `;` at the end

PR #402, opened 2017-07-14
`ut_output_buffer` is now fully abstracted from caller

### LINKED ISSUE #391: Abstract outputbuffer from a reporter class
Make an abstraction of ut_output_reporter and move the logic of operating on outputbuffer and move it to the reporter class.

jgebal:
Resolves #391 
The caller can use reporter to get the output data.

Introduced new abstract type: `ut_output_reporter_base`.
Every reporter based on that type is supposed to use output buffer.
So for every reporter of type `ut_output_reporter_base`, client can call functions:
- get_lines
- lines_to_dbms_output
- get_lines_cursor

Also, the reporters don't need to call parent reporter's method to finalize the reporting.
New finalize event is added.
There is no direct dependency between `ut_runner.run` and `ut_output_buffer` so we may now pass a reporter that doesn't report to output but saves data to table for example.


source/core/types/ut_event_listener.tpb
```diff
@@ -80,15 +78,20 @@ create or replace type body ut_event_listener is
...
-          raise_application_error(ut_utils.gc_invalid_rep_event_name,'Invalid reporting event name - '|| nvl(a_event_name,'NULL'));
         end if;
-      else
-        raise_application_error(ut_utils.gc_invalid_rep_event_time,'Invalid reporting event time - '|| nvl(a_event_timing,'NULL'));
       end if;
     end loop;
 
   end fire_event;
 
+  overriding member procedure fire_event(self in out nocopy ut_event_listener, a_event_name varchar2) is
```
Pazus: I propose to call it `fire on event` to be consistent with other similar procedures

source/core/types/ut_reporter_base.tps
```diff
@@ -60,7 +56,8 @@ create or replace type ut_reporter_base authid current_user as object(
   member procedure after_calling_suite(self in out nocopy ut_reporter_base, a_suite in ut_logical_suite),
 
   -- run hooks continued
-  member procedure after_calling_run (self in out nocopy ut_reporter_base, a_run in ut_run)
+  member procedure after_calling_run (self in out nocopy ut_reporter_base, a_run in ut_run),
+  not instantiable member procedure finalize(self in out nocopy ut_reporter_base)
```
Pazus: Could you add an explaining comment so the new reporter developer understand what he has to implement there if his reporter is not an output reporter

source/core/types/ut_event_listener.tpb
jgebal: ok

source/core/types/ut_reporter_base.tps
jgebal: will do

jgebal: @pazus, or maybe it should have default (empty) implementation, like the rest of the methods?

jgebal:
Closing this PR, will reopen from utPLSQL project feature branch.

PR #405, opened 2017-07-14
Added XPath support for column exclusion in cursor.

### LINKED ISSUE #242: Add ability to exclude columns/attributes for cursor/object/collection comparison
When comparing cursors/object/collections it is often very convenient to be able to skip/ignore some columns/attributes.
The most common scenario would be when the tested data contains create/update timestamp.
This data cannot be compared accurately when it is derived from sysdate/systimestamp.

The `ut_equal` matcher should have 2 additional forms:
- accept additional attribute `a_exclude varchar2` - a_exclude is a comma separated values list of columns/attributes to remove
- accept additional attribute `a_exclude_list ut_varchar2_list` - a_exclude_list is a list of columns/attributes to remove

The implementation could be done by simply removing the XML:
```sql
select deletexml( l_xmltype, '//COLUMN' ) from stream;
```
https://docs.oracle.com/cd/E11882_01/server.112/e41084/functions051.htm#SQLRF06202

jgebal:
Resolves #242 
Added support for attribute exclusion in object/collection (Xpath/names CSV string/names list)

TODO - add documentation.


source/core/ut_utils.pkb
```diff
@@ -328,5 +328,33 @@ create or replace package body ut_utils is
     end if;
   end;
 
+  function to_xpath(a_list varchar2) return varchar2 is
+    l_xpath varchar2(32767) := a_list;
+  begin
+    if l_xpath not like '/%' then
+      l_xpath := '//'||replace(l_xpath,',','|//');
```
jgebal: The CSV is the current behavior and I think we should.
But now that I look at it i think we need to be more clever or clearly state in the documentation that:
You can only use CSV list of column/attribute names to exclude or XPath expression. You cannot mix XPath with CSV.
Also not sure if we should use simple `//` to remove attribute/column anywhere in XML, or should we remove only top level when CSV list is used. Probably top-level only is more predictable and less error-prone.
So the to_xpath would need to change.

source/expectations/data_values/ut_data_value_anydata.tpb
```diff
@@ -16,30 +16,65 @@ create or replace type body ut_data_value_anydata as
...
+  end;
+
+  overriding member function compare_implementation(a_other ut_data_value) return integer is
+    l_self_data  xmltype;
+    l_other_data xmltype;
+    l_other  ut_data_value_anydata;
+    l_result integer;
+    procedure exclude_xpaths(a_xml in out nocopy xmltype, a_xpath varchar2) is
+    begin
+      select case when a_xpath is not null then deletexml( a_xml, a_xpath ) else a_xml end into a_xml from dual;
```
jgebal: good point.

source/expectations/data_values/ut_data_value_anydata.tpb
```diff
@@ -16,30 +16,65 @@ create or replace type body ut_data_value_anydata as
...
+    end;
+  begin
+    if a_other is of (ut_data_value_anydata) then
+      l_other := treat(a_other as ut_data_value_anydata);
+      --needed for 11g xe as it fails on constructing XMLTYPE from null ANYDATA
+      if not self.is_null() and not l_other.is_null() then
+        ut_expectation_processor.set_xml_nls_params();
+        l_self_data := xmltype.createxml(self.data_value);
+        l_other_data := xmltype.createxml(l_other.data_value);
+        exclude_xpaths(l_self_data, coalesce(self.exclude_xpath, l_other.exclude_xpath));
```
jgebal: You actually can pass the exclude list in one place only. I think I've overdone this one-coalesce is not needed.

source/expectations/data_values/ut_data_value_anydata.tpb
```diff
@@ -16,30 +16,65 @@ create or replace type body ut_data_value_anydata as
...
+        end if;
+      end if;
+    else
+      raise value_error;
+    end if;
+    return l_result;
   end;
 
-  overriding member function compare_implementation( a_other ut_data_value ) return integer is
+  final member procedure init(self in out nocopy ut_data_value_anydata, a_value anydata, a_self_type varchar2) is
```
jgebal: I don't want subclasses to override the init procedure but simply use the shared one

tests/ut_expectations/ut.expect.to_equal.anydata.GivesSuccessWithExcludedCollectionAttributes.sql
```diff
@@ -0,0 +1,26 @@
+--Arrange
+declare
+  l_actual   ut_key_value_pairs;
+  l_expected ut_key_value_pairs;
+  l_result   integer;
+  l_results_details ut_expectation_results;
+begin
+--Act
+  l_actual   := ut_key_value_pairs(ut_key_value_pair(key=>'A',value=>'1'), ut_key_value_pair(key=>'B',value=>'2'));
```
jgebal: I don't understand. Do you mean the a_exclude in line 12?

source/expectations/data_values/ut_data_value_anydata.tpb
jgebal: I looked into the code and i think we need it for safety.
We use `order member function compare` to do data comparison.
Therefore, in the `ut_equals` matcher, comparison is done by simply checking `if actual = expected then ...`

We cannot guarantee that we will always use `act = exp` and not `exp = act`.
We should expect the same behaviour regardless of that is the order. So we need the `coalesce` here.
I'll add the same comment into the code.

tests/ut_expectations/ut.expect.to_equal.anydata.GivesSuccessWithExcludedCollectionAttributes.sql
Pazus: Ut_key_value_pair type!

source/expectations/data_values/ut_data_value_anydata.tpb
Pazus: How can they introduce new implementations? Open-close principle. I believe it's not final

source/expectations/data_values/ut_data_value_anydata.tpb
Pazus: Ok then!

source/core/ut_utils.pkb
Pazus: I propose Restrict CSV for xpath. Only collection approach

jgebal: My thinking was:
- Let users use CSV for simplicity, as it's very easy to type CSV list
- let them use list, as it gives more power and can be easily managed programmatically 
I don't think we should move away from CVS approach as it was already introduced in 3.0.0 for cursors.
I need to better understand the difference between CSV and XPath so we can distinguish one from another. That should be quite simple I think, but it's not yet implemented.

PR #417, opened 2017-07-22
Improved performance and stability of access to internal framework tables

jgebal:
When running coverage with several reporters the coverage reported by each of those was a but different.
This should be now resolved as the tmp table is populated only once per coverage run.


source/core/coverage/ut_coverage_helper.pkb
```diff
@@ -16,32 +16,41 @@ create or replace package body ut_coverage_helper is
   limitations under the License.
   */
 
-  g_coverage_id integer;
-  g_develop_mode boolean;
+  g_coverage_id  integer;
+  g_develop_mode boolean := false;
```
Pazus: how about making it not null?

source/core/coverage/ut_coverage_helper.pkb
```diff
@@ -59,17 +68,18 @@ create or replace package body ut_coverage_helper is
   end;
 
   procedure coverage_stop is
-    l_return_code binary_integer;
   begin
     if not g_develop_mode then
```
Pazus: Is it correct to check for develop here? We are stoping coverage regardless of the mode, aren't we?

source/core/coverage/ut_coverage_helper.pkb
```diff
@@ -59,17 +68,18 @@ create or replace package body ut_coverage_helper is
...
+      g_is_started := false;
+      dbms_profiler.stop_profiler();
     end if;
   end;
 
   procedure coverage_stop_develop is
-    l_return_code binary_integer;
   begin
-    l_return_code := dbms_profiler.stop_profiler();
+    g_develop_mode := false;
```
Pazus: I think we should call `coverage_stop` from here not to duplicate code

source/core/coverage/ut_coverage.pkb
```diff
@@ -98,17 +98,46 @@ create or replace package body ut_coverage is
...
+  end;
+
+
   /**
   * Public functions
   */
   procedure coverage_start is
   begin
-    ut_coverage_helper.coverage_start('utPLSQL Code coverage run '||ut_utils.to_string(systimestamp));
+    if not ut_coverage_helper.is_started() then
```
Pazus: we schec the state here and perform the same check in tha package itself.
Maybe we can remove the chack from here and leave it only in `ut_coverage_helper`?

source/core/coverage/ut_coverage_helper.pkb
jgebal: No, not really.
We need to override `coverage_stop` in out unit tests (for now), so that when we test the framework, our "develop-mode" coverage will not stop, if we test execution for one of coverage reporters.

We will be able to remove the develop mode, once we move to testing ut3 with another ut3 installation. 

It needs to stay like this for now, otherwise we will not get a true picture of coverage for out own code.

source/core/coverage/ut_coverage_helper.pkb
jgebal: We can't, as described in previous comment.
The start_develop/stop_develop override and block start/stop of coverage gathering.

PR #424, opened 2017-07-26
Own docker images

Pazus:
Move to use of images from our own repo for builds in Travis.
Added ability to override image tag for a build using variables


jgebal:
how can I access those images?
Did you create a new org/user on docker to host the images?

Pazus:
Look slack chat for details

PR #429, opened 2017-07-28
Implemented reporting to dbms_application_info

### LINKED ISSUE #428: Report activity to dbms_application_info
Report current activity of the framework using dbms_application_info

Pazus:
Resolves #428

### DIFF
--- a/source/core/types/ut_executable.tpb
+++ b/source/core/types/ut_executable.tpb
@@ -76,6 +76,7 @@ create or replace type body ut_executable is
     l_completed_without_errors boolean := true;
     l_start_transaction_id     varchar2(250);
     l_end_transaction_id     varchar2(250);
+    l_client_info            varchar2(200);
     procedure save_dbms_output is
       l_status number;
       l_line varchar2(32767);
@@ -96,6 +97,21 @@ create or replace type body ut_executable is
   begin
     if self.is_defined() then
       l_start_transaction_id := dbms_transaction.local_transaction_id(true);
+      
+      l_client_info := case
+                         when a_item is of(ut_test) then
+                          'Test: '
+                         else
+                          'Suite: '
+                       end || a_item.name || case
+                         when self.associated_event_name = ut_utils.gc_test_execute then
+                          null
+                         else
+                          ' (' || self.associated_event_name || ')'
+                       end;
+      -- report to application_info
+      ut_utils.set_client_info(l_client_info);
+      
       --listener - before call to executable
       a_listener.fire_before_event(self.associated_event_name, a_item);
 
@@ -144,6 +160,7 @@ create or replace type body ut_executable is
       if l_start_transaction_id != l_end_transaction_id or l_end_transaction_id is null then
         a_item.add_transaction_invalidator(self.form_name());
       end if;
+      ut_utils.set_client_info(null);
     end if;
 
     return l_completed_without_errors;
--- a/source/core/types/ut_executable.tpb
+++ b/source/core/types/ut_executable.tpb
@@ -76,7 +76,6 @@ create or replace type body ut_executable is
     l_completed_without_errors boolean := true;
     l_start_transaction_id     varchar2(250);
     l_end_transaction_id     varchar2(250);
-    l_client_info            varchar2(200);
     procedure save_dbms_output is
       l_status number;
       l_line varchar2(32767);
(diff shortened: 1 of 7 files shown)


source/core/types/ut_test.tpb
```diff
@@ -54,6 +54,9 @@ create or replace type body ut_test as
 
     a_listener.fire_before_event(ut_utils.gc_test,self);
     self.start_time := current_timestamp;
+    
+    -- report to application_info
+    ut_utils.set_action(self.object_owner||'.'||self.object_name||'.'||self.name);
```
jgebal: can you move this call to `ut_executable`, so we have visibility into before/after procedures that get executed?

jgebal: I would use: `set_client_info` as it is 64 bytes. We should probably only put `package_name.procedure_name` so that it fits.
```
BEGIN
  DBMS_APPLICATION_INFO.set_module(ut.version, 'before_all');
  DBMS_APPLICATION_INFO.set_client_info(self.object_name||'.'||self.name);
END;
/
```

Pazus:
New behaviour is:

Event|action|client_info
---------|------------|-------
before_all|Suite: <suite_name>|Suite: <suite_name> (before_all)
before_each|Suite: <suite_name>|Test: <test_name> (before_each)
before_test|Suite: <suite_name>|Test: <test_name> (before_test)
test|Suite: <suite_name>|Test: <test_name>
after_test|Suite: <suite_name>|Test: <test_name> (after_test)
after_each|Suite: <suite_name>|Test: <test_name> (after_each)
after_all|Suite: <suite_name>|Suite: <suite_name> (after_all)

source/core/types/ut_executable.tpb
```diff
@@ -96,6 +97,21 @@ create or replace type body ut_executable is
   begin
     if self.is_defined() then
       l_start_transaction_id := dbms_transaction.local_transaction_id(true);
+      
+      l_client_info := case
+                         when a_item is of(ut_test) then
+                          'Test: '
+                         else
+                          'Suite: '
+                       end || a_item.name || case
```
jgebal: I think it would be better give the name from the executed procedure:
`self.object_name||'.'||self.procedure_name`
or even only `self.procedure_name`
This makes it easier to identify the actual procedure running.

tests/ut_test/ut_test.ApplicationInfoOnExecution.sql
```diff
@@ -0,0 +1,149 @@
...
+    dbms_lob.append(l_output,l_output_data(i));
+  end loop;
+  
+  execute immediate 'begin :i := ut_output_tests.gv_before_all_client_info; end;' using out l_client_info;
+  if not nvl(l_client_info = 'Suite: ut_output_tests (before_all)', false) then
+    dbms_output.put_line('Wrong before all text: '||l_client_info);
+    l_result := false;
+  end if;
+  execute immediate 'begin :i := ut_output_tests.gv_before_each_client_info; end;' using out  l_client_info;
+  if not nvl(l_client_info = 'Test: the_test (before_each)', false) then
```
jgebal: > l_client_info = 'Test: the_test (before_each)'

`l_client_info = 'Test: ut_output_tests.before_each (before_each)'`
or
`l_client_info = 'Test: before_each (before_each)'`

source/core/types/ut_executable.tpb
Pazus: but then most likely users will see preparation procedures cause they tend to length more then actually tests. This is why I've chosen such approach

tests/ut_test/ut_test.ApplicationInfoOnExecution.sql
Pazus: 'Test: before_each (before_each)' this way you might be confused, cause its acrually not a test
So then we should use simple `'before_each'` or `'ut_output_tests.before_each'` and nothing else

jgebal: Sounds good to me.
It seems more valuable to be able to identify the before/after procedure invoked that is currently running than the test/suite name that the procedure belongs to, as it helps to track down the slow pieces of code.

Pazus:
Simplified behaviour:

Event|action|client_info
---------|------------|-------
before_all|<suite_name>|<before_all_proc>
before_each|suite_name|<before_each_proc>
before_test|<suite_name>|<before_test_proc>
test|<suite_name>|<test_name>
after_test|<suite_name>|<after_test_proc>
after_each|<suite_name><after_each_proc>
after_all|<suite_name>|<after_all_proc>

Module is always "utPLSQL"

PR #433, opened 2017-08-02
Fixed issue with Sonar Test reporter giving bad reports with suitepath.

### LINKED ISSUE #413: Sonar Unit Test reporter is not working well with suitepath
If a test run contains suitepath, the sonar test execution XML contains something like:
```xml
<testExecutions version="1">
  <file path="equity_pm_database_all">
    <file path="ODS/INSTALL/UNIT_TEST/EQ_PM/PACKAGES/UT_ANALYST_ASSIGN_COV.pkb">
      <testCase name="ut_prc_get_assign_cov" duration="311" >
      </testCase>
    </file>
```
This is not a valid XML

jgebal:
Resolves #413
@pazus, please verify the test for TeamCity test reporter as it seems to be giving suspicious outputs with suitepath too.
Do we need to fix this?

### DIFF
--- a/tests/ut_reporters/ut_teamcity_reporter.ProducesExpectedOutputs.sql
+++ b/tests/ut_reporters/ut_teamcity_reporter.ProducesExpectedOutputs.sql
@@ -3,7 +3,11 @@ declare
   l_output            clob;
   l_expected          varchar2(32767);
 begin
-  l_expected := q'{##teamcity[testSuiteStarted timestamp='%' name='A suite for testing different outcomes from reporters']
+  l_expected := q'{%##teamcity[testSuiteStarted timestamp='%' name='org']
+%##teamcity[testSuiteStarted timestamp='%' name='utplsql']
+%##teamcity[testSuiteStarted timestamp='%' name='utplsql']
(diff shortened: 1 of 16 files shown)


tests/ut_reporters/ut_teamcity_reporter.ProducesExpectedOutputs.sql
```diff
@@ -3,7 +3,11 @@ declare
   l_output            clob;
   l_expected          varchar2(32767);
 begin
-  l_expected := q'{##teamcity[testSuiteStarted timestamp='%' name='A suite for testing different outcomes from reporters']
+  l_expected := q'{%##teamcity[testSuiteStarted timestamp='%' name='org']
+%##teamcity[testSuiteStarted timestamp='%' name='utplsql']
```
jgebal: @Pazus I'm not sure If this would actually be a valid sequence of messages on TeamCity.
Mainly due to the fact that we have suitepath=`org.utplsql.utplsql.test`.
Maybe teamcity suite name should actually show path? Like it is done in XUnit reporter (see the XUnit test below).

PR #435, opened 2017-08-02
Fix problem with identifying annotations when windows newline is used

pesse:
Had this strange behaviour today in a project where I develop on a windows machine using dbForge

### DIFF
--- a/source/core/ut_annotations.pkb
+++ b/source/core/ut_annotations.pkb
@@ -68,7 +68,7 @@ create or replace package body ut_annotations as
       l_comment := a_comments(l_comment_index);
 
       -- strip everything except the annotation itself (spaces and others)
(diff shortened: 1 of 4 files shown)


source/core/ut_annotations.pkb
```diff
@@ -68,7 +68,7 @@ create or replace package body ut_annotations as
       l_comment := a_comments(l_comment_index);
 
       -- strip everything except the annotation itself (spaces and others)
-      l_annotation_str := regexp_substr(l_comment, c_annotation_pattern, 1, 1, modifier => 'i');
+      l_annotation_str := regexp_substr(trim(translate(l_comment, chr(10)||chr(13), ' ')), c_annotation_pattern, 1, 1, modifier => 'i');
```
jgebal: I think that it would be better to put the fix into: `ut_metadata.get_package_spec_source` so that the whole source that we parse have Unix-style end of line.

pesse: Agreed and fixed. Thanks for the hint.

PR #436, opened 2017-08-02
Feature/reporters object model fix

### LINKED ISSUE #391: Abstract outputbuffer from a reporter class

jgebal:
Resolves #391
The caller can use reporter to get the output data.

Introduced new abstract type: ut_output_reporter_base.
Every reporter based on that type is supposed to use output buffer.
So for every reporter of type ut_output_reporter_base, client can call functions:

    get_lines
    lines_to_dbms_output
    get_lines_cursor

Also, the reporters don't need to call parent reporter's method to finalize the reporting.
New finalize event is added.
There is no direct dependency between ut_runner.run and ut_output_buffer so we may now pass a reporter that doesn't report to output but saves data to table for example.


jgebal:
I'm closing this PR for now it requires a bit more work and cleanup

PR #464, opened 2017-08-21
Added `version_compatibility_check` function to ut_runner

### LINKED ISSUE #430: Add ability to check version compatibility
We need a way for the external APIs (like Java) to be able to check if it is compatible, when calling utPLSQL, so that the calling (Java) API can first check if the version that it is depending on is compatible with current version.
That could be done by providing a compatibility check.
`ut_run.check_compatibility_with_version( a_version varchar2)`
The function would return true/false depending on the result of check.
The check would execute the following pseudo-code:
```
begin
  l_result := false;
  l_version = current_version number (major.minor.bugfix);
  if a_version.major = l_version.major
     and (a_version.minor = l_version.minor 
          or a_version.minor = l_version.minor and a_version.bugfix <= l_version.bugfix) then
    l_result := true;
  end if;
  return l_result;
end if;
```
So versions are compatible:
- if major is equal and framework minor >= requested
- if major is equal and minor is equal and framework bugfix >= requested

jgebal:
Resolves #430 
Added version parsing.
Changed value of `gc_version` variable in `ut_utils`


source/core/ut_utils.pkb
```diff
@@ -384,5 +384,18 @@ create or replace package body ut_utils is
     commit;
   end;
 
+  function to_version(p_version_no varchar2) return t_version is
```
pesse: IMO this function should raise an exception with proper message if an illegal version-no string is provided.

source/api/ut_runner.pkb
```diff
@@ -43,6 +43,19 @@ create or replace package body ut_runner is
     return ut_utils.gc_version;
   end;
 
+  function version_compatibility_check( a_requested varchar2, a_current varchar2 := null ) return boolean is
```
pesse: For boolean is not a SQL type but only PL/SQL (I guess it's still working that way) it might be more complicated to implement the Java request to this function, cause we can't just `select version_compatibility_check(?) from dual`
What do you think about return 0/1? Or adding another function, wrapping the boolean one and returning 0/1?

test/ut_runner/test_ut_runner.pkb
```diff
@@ -0,0 +1,22 @@
...
+    ut.expect( ut3.ut_runner.version_compatibility_check('v3.0.0.0','v2.0.0.0') ).to_be_false;
+    ut.expect( ut3.ut_runner.version_compatibility_check('v3.0.0.0','v4.0.0.0') ).to_be_false;
+  end;
+  procedure version_comp_check_ignore is
+  begin
+    ut.expect( ut3.ut_runner.version_compatibility_check('v3.0.0.123','v3.0.0.0') ).to_be_true;
+    ut.expect( ut3.ut_runner.version_compatibility_check('v3.0.0.0','v3.0.0.123') ).to_be_true;
+    ut.expect( ut3.ut_runner.version_compatibility_check('v3.0.0','v3.0.0.0') ).to_be_true;
+  end;
+
```
pesse: There should be tests to ensure version_compatibility_check behaves correctly (exception) when providing illegal version strings

source/api/ut_runner.pkb
jgebal: Interesting observation.
We already have bool ->init and int->bool so the change will be quite simple.
I'm just wondering how would the select statement behave with overloaded functions (with boolan and integer return type) - will need to check

test/ut_runner/test_ut_runner.pkb
jgebal: good point

source/api/ut_runner.pkb
Pazus: You can't overload a funtion with a version which differes ony in return type.

Pazus:
I'm ok with the changes so lets wait for @pesse reviewal.

source/core/ut_utils.pks
```diff
@@ -235,5 +248,14 @@ create or replace package ut_utils authid definer is
 
   procedure cleanup_temp_tables;
 
+  /**
+  * Converts version string into version record
+  *
+  * @param    a_version_no string representation of version in format vX.X.X.X where X is a positive integer
+  * @return   t_version    record with up to four positive numbers containing version
+  * @throws   20010        if passed version string is not matching version pattern
```
pesse: @throws 20214, not 20010

source/api/ut_runner.pks
```diff
@@ -20,6 +20,19 @@ create or replace package ut_runner authid current_user is
...
   /**
+  * Check if version is compatible with another version (by default the current framework version)
+  * Version is compatible if:
+  *   a_current.major = a_requested.major
+  *   a_requested.minor < a_current.minor or a_requested.minor = a_current.minor and a_requested.bugfix <= a_current.bugfix
+  *
+  * @param     a_requested requested utPLSQL version string
+  * @param     a_current   current   utPLSQL version string, if null is passed, defaults to current framework version
+  * @return    1/0         1-true, 0-false
+  * @exception 20010       if passed version string is not matching version pattern
```
pesse: @throws 20214, not 20010

PR #465, opened 2017-08-22
Added tests from old tests to new approach

pesse:
Added test for Windows-Newline in annotations and annotations with very long names (oracle 12c allows that) to new testing approach

### DIFF
--- a/test/install_tests.sql
+++ b/test/install_tests.sql
@@ -12,7 +12,6 @@
 @core.pkb
 @ut_utils/test_ut_utils.pkb
 @ut_annotations/test_annotations.pkb
-show errors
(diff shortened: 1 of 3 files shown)


test/install_tests.sql
```diff
@@ -12,6 +12,7 @@
 @core.pkb
 @ut_utils/test_ut_utils.pkb
 @ut_annotations/test_annotations.pkb
+show errors
```
jgebal: can you move the show errors to the end of script?

pesse: Doesn't work at the end of the script (there is already one) - but I can remove it again of course.

PR #466, opened 2017-08-22
Removed autonomous_transaction from cursor init

### LINKED ISSUE #462: Error when trying to use a Global Temporary Table
Hi - If the procedure that I am testing uses a Global Temporary Table then I get an error when I try to test it.  I get the same error if I try to use a GTT in the test suite to store expected results (which I would then compare to the actual table).

**ORA-19202: Error occurred in XML processing
ORA-08103: object no longer exists
ORA-06512: at "UT3.UT_DATA_VALUE_REFCURSOR", line 105
ORA-06512: at "UT3.UT_DATA_VALUE_REFCURSOR", line 30
ORA-06512: at "UT3.UT_EQUAL", line 70
ORA-06512: at "UT3.UT_EXPECTATION_REFCURSOR", line 21**

When I change these to regular tables the test suite runs fine (without changing any code).

Any idea about this error/bug?

Thanks,
Alan Rosenthal
alan.h.rosenthal@gmail.com

P.S.  I LOVE utPLSQL !!!!!!!!!!!!!

pesse:
Fixes #462
Added additional test for cursor comparison on temporary tables


pesse:
Removing just the autonomous transaction thing causes an implicit rollback via set_xml_nls_params (i guess) - this issue has to be investigated more

test/ut_expectations/test_expectations_cursor.pkb
```diff
@@ -0,0 +1,42 @@
...
+    v_expected(1).value := 'Test-entry';
+
+    -- insert test-data into temporary table
+    execute immediate 'insert into gtt_test_table ( value ) values ( ''Test-entry'' )';
+
+    -- test 
+    open c_expected for select * from table(v_expected);
+    open c_actual for 'select * from gtt_test_table';
+
+    ut.expect( c_actual ).to_equal( a_expected => c_expected );
```
jgebal: `ut3.ut.expect` I guess

jgebal: Actually that should be something like:
```sql
begin
  --Arrange
  execute immediate 'insert into gtt_test_table ( value ) values ( ''Test-entry'' )';
  open c_expected for select 'Test-entry' as value from dual;
  open c_actual for 'select * from gtt_test_table';
  --Act - execute the expectation on cursor opened on GTT
  ut3.ut.expect( c_actual ).to_equal( a_expected => c_expected );
  --Assert - check that expectationm was executed successfully
  ut.expect(ut3.ut_expectation_processor.get_status()).to_equal(ut3.ut_utils.tr_success);
  --Cleanup
  ut3.ut_expectation_processor.clear_expectations();
end;
```

source/create_synonyms_and_grants_for_public.sql
```diff
@@ -88,6 +88,7 @@ create public synonym ut_runner for &&ut3_owner..ut_runner;
 create public synonym ut_teamcity_reporter for &&ut3_owner..ut_teamcity_reporter;
 create public synonym ut_xunit_reporter for &&ut3_owner..ut_xunit_reporter;
 create public synonym ut_documentation_reporter for &&ut3_owner..ut_documentation_reporter;
+create public synonym ut_ansiconsole_helper for &&ut3_owner..ut_ansiconsole_helper;
```
jgebal: I would remove this synonym. ut_ansiconsole_helper is `internal` code and we don't need to make it publicly available.
Not exposing it publicly gives us more freedom to change internals of the project.

source/create_synonyms_and_grants_for_user.sql
```diff
@@ -108,6 +108,7 @@ create or replace synonym &ut3_user..ut_runner for &&ut3_owner..ut_runner;
 create or replace synonym &ut3_user..ut_teamcity_reporter for &&ut3_owner..ut_teamcity_reporter;
 create or replace synonym &ut3_user..ut_xunit_reporter for &&ut3_owner..ut_xunit_reporter;
 create or replace synonym &ut3_user..ut_documentation_reporter for &&ut3_owner..ut_documentation_reporter;
+create or replace synonym &ut3_user..ut_ansiconsole_helper for &&ut3_owner..ut_ansiconsole_helper;
```
jgebal: I would remove this synonym. ut_ansiconsole_helper is `internal` code and we don't need to make it publicly available.
Not exposing it publicly gives us more freedom to change internals of the project.

PR #486, opened 2017-09-28
Added example for custom expectation-fail message in docs

### LINKED ISSUE #485: Add possibility to provide custom error messages to expect
Sometimes you have situations where you dynamically test things on several different tables. 
Example: You have a bunch of tables and an archive-functionality for them and you want to test if the things you put into live-tables are removed from live-tables and present in archive-tables:

````
procedure test_data_existance( i_tableName varchar2 ) 
  as
    v_count_real integer;
    v_count_archive integer;
  begin
    
    execute immediate 'select count(*) from ' || i_tablename || '' into v_count_real;
    execute immediate 'select count(*) from ' || i_tablename || '_archive' into v_count_archive;

    ut.expect( v_count_archive ).to_( equal(1) );
    ut.expect( v_count_real ).to_( equal(0) );

  end;

 procedure test_ARCHIVE_DATA
  as
  begin
    -- Arrange
   -- insert several data into real-tables here

    -- Act
    package_to_test.archive_data();

    -- Assert
    test_data_existance('TABLE_A');
    test_data_existance('TABLE_B');
    test_data_existance('TABLE_C');
    test_data_existance('TABLE_D');
end;
````
If something fails I have no clue for which of my dynamic tests it failed.
It would therefore be great to have the possibility to add a dynamic error message to the expect object. Maybe somthing like 

`ut.expect( <actual> ).to_( <matcher> ).raise_( <custom error message if fails> )`

### PR DESCRIPTION
**Pull request by pesse**
Fixes #485

### DIFF
--- a/docs/userguide/expectations.md
+++ b/docs/userguide/expectations.md
@@ -6,17 +6,18 @@ To achieve that, we use a combination of expectation and matcher to perform the 
 Example of a unit test procedure body.
 ```sql
 begin
-  ut.expect( 'the tested value' ).to_( equal('the expected value') );
+  ut.expect( 'the tested value', 'optional custom failure message' ).to_( equal('the expected value') );
 end;
 ```
 
 Expectation is a set of the expected value(s), actual values(s) and the matcher(s) to run on those values.
+You can also add a custom failure message for an expectation.
 
 Matcher defines the comparison operation to be performed on expected and actual values.
 Pseudo-code:
 ```
-  ut.expect( a_actual {data-type} ).to_( {matcher} );
-  ut.expect( a_actual {data-type} ).not_to( {matcher} );
+  ut.expect( a_actual {data-type} [, a_message {varchar2}] ).to_( {matcher} );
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: docs/userguide/expectations.md [hunk 1]
```diff
@@ -471,3 +471,55 @@ end;
 ```
 Since NULL is neither *true* nor *not true*, both expectations will report failure.
 
+# Provide a custom error message
```
jgebal: To be consistent with the rest of document please change that to: Adding custom failure message

docs/userguide/expectations.md
```diff
@@ -471,3 +471,55 @@ end;
...
+Expectations allow you to provide a custom error message as second argument:
+````sql
+ut.expect( a_actual {data-type}, 'custom error message if expectation fails' ).to_{matcher};
```
jgebal: Can you replace the 'error message' with 'failure message' or just 'custom message'.
There is a difference between error and failure as we define errored test as a test procedure that raised an unhandled exception.

PR #492, opened 2017-10-14
Adding cache and restructuring annotations 

### LINKED ISSUE #458: Test execution takes ages for large projects with thousands of packages
For large projects having thousands of packages in tested schema (like OEBS apps schema) one simple test takes about 20 minutes. The test engine tries to find all the annotations in ALL_SOURCE view wich is time consuming.
Can we specify optional test packages name prefix before execution (or somewhere in configuration) to reduce the scan efforts?

jgebal:
* Removed duplicate `old_tests` for ut_annotations - the functionality is covered with new tests.
* Removed parsing of annotation text (params) from ut_annotations as annotation text can be different depending on use case and it's not parsers responsibility to support all use-cases for annotation text.
* Moved annotations code to separate directory
* Added filtering of dba_source by test containing annotations.
* Added cardinality hint with scaling.
* Changed naming of procedures and variables.
* Refactored annotations API to use collections of database objects as output results.
* Reworked `ut_suite_manager` to operate on sorted collection for building ut_suite and it's tests.
* Added cache mechanism to `ut_annotation_parser`.
* Changed the way schema is scanned, so that all annotations/sources are scanned with single execution of a query.
* TODO - change the way annotations are parsed, so that we can scan for "floating" annotations.
* Annotations that are on top of suite, but associated with procedure, are not valid suite-level annotations (This is why the package  `old_tests/helpers/test_reporters` was updated).
* Refactored annotations tests

Resolves #458


jgebal:
I see now that it's missing unit tests.
`ut_annotation_manager` and `ut_annotation_cache_manager` need to be unit-tested.
I could use some help with that as I will not be able to spend much more time on the project this week.

Pazus:
Could you add a `populate_cache` procedure to persist the sources without running tests and a method to clear the cache.

jgebal:
@Pazus I'm working on it now. Also refactoring the code a bit - better modularity and segregation of responsibility.

jgebal:
@pazus
The ability to build / flush cache is now public in ut_runner.

PR #496, opened 2017-10-17
Do not open new transaction when getting called.

### LINKED ISSUE #441: Finish ut.run without active transaction
When tests are executed on the first savepoint creation transaction is initiated. After rollback to that savepoint issued the transaction remains active even though it contains no changes and IDEs offer to rollback/commit changes. 

I propose to check if no transaction is active at the start of execution and rollback (complete rollback) at the end of the run.

jgebal:
With this change all calls to `ut.run` are in autonomous transaction.
This way, utPLSQL does not create a new transaction if a transaction was not opened in the calling session.

Resolves #441


source/api/ut.pkb
```diff
@@ -259,55 +259,55 @@ create or replace package body ut is
...
     a_coverage_schemes ut_varchar2_list := null, a_source_file_mappings ut_file_mappings := null, a_test_file_mappings ut_file_mappings := null,
     a_include_objects ut_varchar2_list := null, a_exclude_objects ut_varchar2_list := null
   ) is
+    l_reporter  ut_reporter_base := coalesce(a_reporter, ut_documentation_reporter());
   begin
-    ut.run(
-      ut_varchar2_list(sys_context('userenv', 'current_schema')), a_reporter, a_color_console,
-      a_coverage_schemes, a_source_file_mappings, a_test_file_mappings, a_include_objects, a_exclude_objects
+    run_autonomous(
+      a_paths, l_reporter, ut_utils.boolean_to_int(a_color_console), a_coverage_schemes, a_source_file_mappings, a_test_file_mappings,
```
Pazus: Change argument to `to _a_reporter`

Pazus: My bad, skip it.

PR #512, opened 2017-11-12
Fixed coverage filtering issue.

### LINKED ISSUE #511: No coverage data, when running with a_include_objects from different schema
When I execute the tests as below:
```
alter session set current_schema=other_schema;

select *
  from table(
    ut.run(
      user, 
      ut_coverage_html_reporter(),
      a_include_objects => ut_varchar2_list( user||'.betwnstr')
    )
  );
```
The coverage report is empty.
I need to explicitly specify `a_coverage_schemes=>ut_varchar2_list( user )` as a parameter, even though the owner of the object is included as part of the `a_include_objects` parameter.

The behavior should be:

 a_include_objects | a_coverage_schemes | a_source_file_mappings | coverage filtering |
-|-|-|-|
 null | null | null | all objects (except tests) in current_schema|
 specified | null | null | all objects (except tests) from the list specified (as per schemes on the object list) |
 null | specified | null | all objects in schemes specified (except unit tests)  |
 null | null | specified | all objects (except tests) from the file mapping list (as per schemes on the list) |

It doesn't really make sense to support combinations of those 3 parameters.
Those parameters should be mutually exclusive.
So users should only provide one of parameters not a combination of those.

**Current implementation sets value of `a_coverage_schemes` to be CURRENT_SCHEMA, if not value was provided. This makes the `a_include_objects` filter useless without specifying schema names.**

jgebal:
Coverage can now be gathered with `a_include_objects` parameter
Resolves #511

### DIFF
--- a/source/api/ut_runner.pks
+++ b/source/api/ut_runner.pks
@@ -33,28 +33,31 @@ create or replace package ut_runner authid current_user is
   function version_compatibility_check( a_requested varchar2, a_current varchar2 := null ) return integer;
 
   /**
-  * Run suites/tests by path
-  * Accepts value of the following formats:
-  * schema - executes all suites in the schema
-  * schema:suite1[.suite2] - executes all items of suite1 (suite2) in the schema.
+  * Execute specified suites/tests by paths
+  * @param a_paths                  list of schemes, packages, procedures or suite-paths to execute
+  * @param a_reporters              list of reporter objects (formats) to use for reporting
+  * @param a_color_console          true/false - should the console format reporters use ANSI color tags
+  * @param a_coverage_schemes       list of database schemes to include in coverage
+  * @param a_source_file_mappings   list of project source files mapped to DB objects that coverage should be reported on
+  * @param a_test_file_mappings     list of project test files mapped to DB objects that test results should be reported on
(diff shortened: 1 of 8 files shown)


source/api/ut_runner.pks
```diff
@@ -33,31 +33,34 @@ create or replace package ut_runner authid current_user is
   function version_compatibility_check( a_requested varchar2, a_current varchar2 := null ) return integer;
 
   /**
-  * Run suites/tests by path
-  * Accepts value of the following formats:
-  * schema - executes all suites in the schema
-  * schema:suite1[.suite2] - executes all items of suite1 (suite2) in the schema.
+  * Execute specified suites/tests by paths
+  * @param a_paths                  list of schemes, packages, procedures or suite-paths to execute
```
Pazus: The “at” sign will cause problems in some environments (for example PL/SQL Developer)
Do you mind it changing to “%”?

pesse: Oh! Didn't know. I always use at in description because it works with plsql doc...

jgebal:
If possible I would rather keep the @ sign on comments.
The will be used to generate documentation. 
Both SQLDevelope a

jgebal:
oh. I didn't mean to close it - reopening.

PR #519, opened 2017-11-23
Added me to the contributors

pesse:


pom.xml
```diff
@@ -0,0 +1,62 @@
+<!--Used to obtain ojdbc8 and orai18n files that are needed for client-->
```
jgebal: Maybe the pom file should also exist in .travis, as it's only used for CI?
I'd like to avoid confusion around the project using maven.
The utPLSQL project is not a maven project and the existence of `pom.xml` in project root is suggesting that it is.

.travis.yml
```diff
@@ -69,11 +73,15 @@ before_install:
   #- curl -LOk $(curl --silent https://api.github.com/repos/utPLSQL/utPLSQL/releases/latest | awk '/browser_download_url/ { print $2 }' | grep ".zip" | sed 's/"//g')
   - git clone --depth=1 --branch=${SELFTESTING_BRANCH:-master} https://github.com/utPLSQL/utPLSQL.git $UTPLSQL_DIR
   # download utPLSQL-cli develop
-  - curl -Lk -o utPLSQL-cli.zip https://bintray.com/viniciusam/utPLSQL-cli/download_file?file_path=${UTPLSQL_CLI_FILE}
+  - curl -Lk -o utPLSQL-cli.zip https://github.com/utPLSQL/utPLSQL-cli/releases/download/v3.0.4/${UTPLSQL_CLI_FILE}
 
 install:
   #- unzip utPLSQL.zip
   - unzip utPLSQL-cli.zip && chmod -R u+x utPLSQL-cli
+  - cp .travis/settings.xml $MAVEN_CFG/settings.xml
```
jgebal: can we move this to `maven_cfg.sh`?

.travis.yml
```diff
@@ -69,11 +73,15 @@ before_install:
...
+  # Install maven wagon
+  - bash .travis/maven_cfg.sh
+  - mvn dependency:copy-dependencies -DoutputDirectory=utPLSQL-cli/lib
```
jgebal: this could also be in `maven_cfg.sh`

.travis.yml
```diff
@@ -69,11 +73,15 @@ before_install:
   #- curl -LOk $(curl --silent https://api.github.com/repos/utPLSQL/utPLSQL/releases/latest | awk '/browser_download_url/ { print $2 }' | grep ".zip" | sed 's/"//g')
   - git clone --depth=1 --branch=${SELFTESTING_BRANCH:-master} https://github.com/utPLSQL/utPLSQL.git $UTPLSQL_DIR
   # download utPLSQL-cli develop
-  - curl -Lk -o utPLSQL-cli.zip https://bintray.com/viniciusam/utPLSQL-cli/download_file?file_path=${UTPLSQL_CLI_FILE}
+  - curl -Lk -o utPLSQL-cli.zip https://github.com/utPLSQL/utPLSQL-cli/releases/download/v3.0.4/${UTPLSQL_CLI_FILE}
```
jgebal: File name will not change so no need to put it as variable.
Better to have the version number as variable - it will be subject to change.

Issue #454, opened 2017-08-10
Add ability to report only the rows that are different


jgebal:
## Ability to report only the rows that are different
---------------------
Currently we simply display first N rows form both cursors if the data is not matching.

This is not very useful for developers, specially when dealing with more than few columns and few rows of data.

`ut_data_value_refcursor` should implement some kind of diff report, that would:
- only report on rows that are actually different
- display mismatching rows side by side or one under another so we can see one expected row, one actual etc.

As an idea, when running the `compare_implementation` method, we could save the results (or data-keys) of both compared cursors into separate diff table so that we can then simply query diff table joined with cursor data table (twice) to get differences and the diff data.

It would be worth doing a bit of research on was to do Diff.
There is a nice open-source algorithm for text-based diff:
https://neil.fraser.name/software/diff_match_patch/svn/trunk/demos/demo_diff.html
https://github.com/curran/google-diff-match-patch

RSpec and git command line have a line by line diff implementation too, that could be more suitable here.

mikechip1980:
as far as I understand, the comparison now is implemented by transformation Cursor to XML.
In case of difference we can see only first several rows of the source XML which useless for subsequent analisys. I absolutely agree that this feature should be changed and the output with the difference (like in other Expectations API) would be much more useful.
As variant, XMLDIFF oracle function could be used
https://docs.oracle.com/cd/B28359_01/server.111/b28286/functions229.htm#SQLRF20025
It has 2 XMLs as input parameters for comparison and output XML with detail difference information.

jgebal:
Xmldiff would be a good starting point.
The outcomes of diff are however ... well ... ugly.
We would need to transform that into something human-readable.
It's simple enough if compared cursors are flat rows and we could simply display rows that differ. It gets tricky when you have a complex structure like cursor returning object type, collection or nested cursor.
Perhaps that should be handled one by one.
First figure how to display diff for a flat row, then for object, then for collection/nested cursor

mikechip1980:
here is useful link about XML comparison, several variants and some analysis 
https://www.google.ru/amp/s/ellebaek.wordpress.com/2011/02/01/comparing-xml-in-oracle/amp/

mathewbutler:
We should stick with current XML and tools approach but also test for large comparisons - millions of rows to verify performance.

PR #527, opened 2017-11-29
Improved cursor comparison - only showing different lines

pesse:
At least partly fixes https://github.com/utPLSQL/utPLSQL/issues/454


source/expectations/data_values/ut_cursor_data_diff.sql
```diff
@@ -0,0 +1,4 @@
+create global temporary table ut_cursor_data_diff(
```
jgebal: comments in tables are allowed but no empty lines.

source/expectations/data_values/ut_data_value_refcursor.tpb
```diff
@@ -138,17 +138,17 @@ create or replace type body ut_data_value_refcursor as
...
                      ucd.row_no
                 from ut_cursor_data ucd where ucd.cursor_data_guid = self.data_value) exp
         full outer join (select case when l_xpath is not null then deletexml( ucd.row_data, l_xpath ) else ucd.row_data end as row_data,
                                 ucd.row_no
                            from ut_cursor_data ucd where ucd.cursor_data_guid = l_other.data_value) act
          on (exp.row_no = act.row_no)
-       where nvl(dbms_lob.compare(xmlserialize( content exp.row_data no indent), xmlserialize( content act.row_data no indent)),1) != 0
-         and rownum <= 1;
+       where nvl(dbms_lob.compare(xmlserialize( content exp.row_data no indent), xmlserialize( content act.row_data no indent)),1) != 0;
+      select count(1) into l_result from ut_cursor_data_comp where rownum <= 1;
```
Pazus: You can use `SQL%ROWCOUNT` instead of selecting rows number after select.

jgebal:
Can you add tests, so are covered for regression?
Succeeds when cursors are same.
Succeeds bot when cursors are empty.
Fails and returns rows from `actual` that are not in `expected`
Fails and returns rows from `expected` that are not in `actual`

jgebal:
My concern is that there is no limit on the data returned.
If the cursor will contain 100 cols and 10M rows, it will cause some serious issues.

Pazus:
Good point, Jacek. We still need a limit...

source/expectations/data_values/ut_data_value_refcursor.tpb
```diff
@@ -138,17 +138,17 @@ create or replace type body ut_data_value_refcursor as
...
         from (select case when l_xpath is not null then deletexml( ucd.row_data, l_xpath ) else ucd.row_data end as row_data,
                      ucd.row_no
                 from ut_cursor_data ucd where ucd.cursor_data_guid = self.data_value) exp
         full outer join (select case when l_xpath is not null then deletexml( ucd.row_data, l_xpath ) else ucd.row_data end as row_data,
                                 ucd.row_no
                            from ut_cursor_data ucd where ucd.cursor_data_guid = l_other.data_value) act
          on (exp.row_no = act.row_no)
-       where nvl(dbms_lob.compare(xmlserialize( content exp.row_data no indent), xmlserialize( content act.row_data no indent)),1) != 0
-         and rownum <= 1;
+       where nvl(dbms_lob.compare(xmlserialize( content exp.row_data no indent), xmlserialize( content act.row_data no indent)),1) != 0;
```
jgebal: I would add limit here (in the insert statement), so that we only process first 50 - 100 differences.
Why 50 - 100?
10 seems not enough, but anything more than 50 - 100 doesn't bring much value anyway.
If you have so may differences, you're in serious trouble anyway.

If first 100 rows doesn't help you identify what the problem is - your test is not distilled properly and you're testing all in one go probably.

mathewbutler:
I said a long time ago that I’d look at an alternative implementation. The best I’ve done is to put together a simple SQL implementation of a compare. The below does two full table scans to identify differences. It’s an asktom special - forget the name of the person on the three that worked through the problem to come up with this. 

select c1, c2, c3, 
  2         count(src1) CNT1, 
  3          count(src2) CNT2
  4     from 
  5   ( select a.*, 
  6          1 src1, 
  7          to_number(null) src2 
  8       from a
  9      union all
  10    select b.*, 
  11         to_number(null) src1, 
  12         2 src2 
  13      from b
  14   )
  15  group by c1,c2,c3
  16 having count(src1) <> count(src2)

My idea was to use this and dynamically produce the query. With testing to see if special handling needed for all data types.

Laptop times for comparing two 1M tables were .8 seconds and low resource usage.  

Leaving this here, as above forms the basis of a POC. We said that we wouldn’t change the implementation unless there were clear benefits or new requirements. 

Cheers.

jgebal:
@mathewbutler Agreed.
We can always do this for simple datatypes, on top of what we do.

That is, if not clob, blob, xmltype etc - use plain SQL compare, else do to XML and compare CLOBs.

For now, we definitely want to improve reporting side, so that we show rows that differ instead of first 'x' rows from data-set.
I think it's a nice step forward.

source/expectations/data_values/ut_data_value_refcursor.tpb
jgebal: Second thought on this idea.
Compare all rows.
Remember toe count of:
- rows in A
- rows in B
- differences
Report first `x` differences (having `x` 50 .. 100)
Report also number of rows in `A`, number of rows in `B` and total number of differences.

That will give much more information about failed assertions,

pesse: Had something similar in mind myself - would definitely help to output exact number of mismatches because we got them anyways...

pesse:
Seems the changes broke the exclude-possibility of columns. Might take me some time to get it sorted out.

PR #540, opened 2017-12-25
Fixed coverage reporting for labelled `end loop` block.

### LINKED ISSUE #539: Code coverage reports covered lines as missed lines
Code coverage reports the following lines always as "not covered":

- declaration of a private package variable/constant
- ```END LOOP``` of a FOR-LOOP statement

These lines should be treated as irrelevant lines from my point of view. I've attached a coverage report to visualise the problem.

[coverage.html.zip](https://github.com/utPLSQL/utPLSQL/files/1586027/coverage.html.zip)

jgebal:
Resolves #539

### DIFF
--- a/source/core/coverage/ut_coverage.pkb
+++ b/source/core/coverage/ut_coverage.pkb
@@ -67,7 +67,7 @@ create or replace package body ut_coverage is
                      ) and
                      regexp_like(
                         s.text,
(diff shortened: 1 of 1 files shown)


jgebal:
the change from `\s` to `[\t ]` is just for purity as `\s` includes newline character.
We run regex line by line so it should not really matter but was easier to verify the regex without newlines included in search pattern.

jgebal:
Still need to add unit tests to cover the filtering of false-negatives.

PR #556, opened 2018-01-13
When a closed cursor is passed to expect, it throws a exception

### LINKED ISSUE #548: Behavior of ut.expect(closed_cursor)
Currently, when `ut.expect` is executed with a cursor that was closed, no exception is thrown.

I think we should change that behavior as such cursor is invalid and cannot be processed.

I would prefer the call to `ut.expect(closed_cursor) to raise exception.
We do not hide exceptions raised while fetching from cursor (like divisor_equal_to_zero) so I suppose we should not hide that we cannot execute expectation on a closed cursor.

Any thoughts on this? Should we change that behavior?

AlexisGaldamez:
Resolves #548


AlexisGaldamez:
Yes I'll do it, I sent it because Jacek told he want to see how I did it

PR #564, opened 2018-01-28
New throws annotation

### LINKED ISSUE #490: Add --%throws annotation
That way we can write code:
```sql
begin
   a_failing_procedure();
   ut.fail(-20000);
exception
  when other then
    ut.expect(sqlcode).to_equal(-20000);
end;
```
If the `a_failing_procedure();` call returns without exception, the `ut.fail(-20000)` would report this.
The tests when it's called would report:
`Expected exception ORA-20000, but nothing was raised`

Currently, we need to code around it a bit. The way I do it is:
```sql
begin
   a_failing_procedure();
   ut.fail('Expected exception ORA-20000 but nothing was raised');
exception
  when other then
    ut.expect(sqlcode).to_equal(-20000);
end;
```

AlexisGaldamez:
Draft for the new characteristic that will allow the user to write in the annotations the exceptions that the test musth throws.
The syntax for the new characteristic is:
`create or replace package test_stuff is`
`  --%suite`

  `--%test(Raises an exception on closed cursor)`
  `--%throws(-20145)`
  `exception_when_closed_cursor;`
`end;`

Resolves #490


source/core/types/ut_executable.tpb
```diff
@@ -59,23 +60,27 @@ create or replace type body ut_executable is
     return ut_metadata.form_name(owner_name, object_name, procedure_name);
   end;
 
-  member procedure do_execute(self in out nocopy ut_executable, a_item in out nocopy ut_suite_item, a_listener in out nocopy ut_event_listener_base) is
+  member procedure do_execute(self in out nocopy ut_executable, a_item in out nocopy ut_suite_item, a_listener in out nocopy ut_event_listener_base,
```
jgebal: I'm thinking more and more if that functionality (checking for exceptions) should be part of the generic ut_executable, or should we build a separate `ut_executable_test` to clearly separate additional functionality that is dedicated for tests only, not for before/after blocks.
So the `ut_executable_test` could call the base ut_executable.do_execute, and then invoke the custom logic to check fo rexceptions.
In suite builder, we would construct `ut_executable_test` only for the TEST block, and for all the others, we would still use ut_executable

test/core/annotations/test_annot_throws_exception.pkb
```diff
@@ -0,0 +1,76 @@
+create or replace package body test_annot_throws_exception
```
jgebal: Can you add negative scenario tests for:
--%throws - with no parameters
--%throws( some bad and dummy value )
--%throws(-1234,+1234,some bad value)
--%throws(, .. , )
So we know what it does exactly

source/core/types/ut_executable.tpb
AlexisGaldamez: I like this approach. I think it's better create a ut_executable_test class

source/core/types/ut_executable.tpb
```diff
@@ -93,6 +98,25 @@ create or replace type body ut_executable is
...
+    begin
+      if a_error_stack is null then
+        l_fail_message := 'Expected exceptions '||a_expected_error_codes||' but nothing was raised.';
+      else
+        l_actual_error_no := regexp_substr(a_error_stack, '^ORA(-[0-9]+)',subexpression=>1);
+
+        if a_expected_error_codes = l_actual_error_no then
+          l_fail_message := null;
+        else
+          l_fail_message := 'Expected exceptions '''||a_expected_error_codes||''' got '''||l_actual_error_no||'''.';
```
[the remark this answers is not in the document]
AlexisGaldamez: I'm agree, that message it's better

test/core/annotations/test_annot_throws_exception.pkb
AlexisGaldamez: I'll add them

AlexisGaldamez:
About the documentation, when the feature is finished I'll add examples about how to use it to show to the users the new syntaxis @jgebal

source/core/types/ut_executable.tpb
```diff
@@ -93,6 +98,25 @@ create or replace type body ut_executable is
...
+    function build_failed_expec_errnum(a_error_stack in varchar2, a_expected_error_codes in varchar) return varchar is
+      l_actual_error_no integer;
+      l_fail_message varchar2(4000);
+    begin
+      if a_error_stack is null then
+        l_fail_message := 'Expected exceptions '||a_expected_error_codes||' but nothing was raised.';
+      else
+        l_actual_error_no := regexp_substr(a_error_stack, '^ORA(-[0-9]+)',subexpression=>1);
+
+        if a_expected_error_codes = l_actual_error_no then
```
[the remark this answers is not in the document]
AlexisGaldamez: Yes ,maybe It would be nice to create a new function like this *ut_utils.trim_list_elements* which return the elements already trimed

source/core/ut_suite_builder.pkb
```diff
@@ -41,12 +41,27 @@ create or replace package body ut_suite_builder is
...
     end;
+
+    function exception_numbers_list(a_exception_numbers in varchar2) return ut_varchar2_list is
+      l_exception_number_list ut_varchar2_list;
+      l_regexp_for_excep_nums varchar2(30) := '^-?[[:digit:]]{1,5}$';
+    begin
+      /*the a_expected_error_codes is converted to a ut_varchar2_list after that is trimmed and filtered to left only valid exception numbers*/
+      l_exception_number_list := ut_utils.string_to_table(a_exception_numbers, ',', 'Y');
+      l_exception_number_list := ut_utils.trim_list_elements(l_exception_number_list, '[:space:]');
+      l_exception_number_list := ut_utils.filter_list(l_exception_number_list, l_regexp_for_excep_nums);
```
jgebal: Really nice. I love the names of procedures - makes it clear to understand what's going on.

source/core/ut_utils.pkb
```diff
@@ -492,5 +492,37 @@ procedure append_to_clob(a_src_clob in out nocopy clob, a_clob_table t_clob_tab,
...
 
+  function trim_list_elements(a_list IN ut_varchar2_list, a_characters_to_trim in varchar2) return ut_varchar2_list is
+    l_trimmed_list ut_varchar2_list := ut_varchar2_list();
+    l_index integer;
+  begin
+    l_index := a_list.first;
+
+    while (l_index is not null) loop
+      l_trimmed_list.extend;
+      l_trimmed_list(l_trimmed_list.count) := regexp_replace(a_list(l_index), '(^['||a_characters_to_trim||']*)|(['||a_characters_to_trim||']*$)');
```
jgebal: I'd build the filter outside of loop into a `l_fiter` variable. Without that, you're concatenating inside a loop every time the loop gets executed.

test/core/annotations/test_annot_throws_exception.pkb
jgebal: This looks great!

test/core/test_ut_utils.pkb
```diff
@@ -293,6 +293,26 @@ end;';
     ut.expect(l_version.build).to_equal(333);
   end;
 
+  procedure test_trim_list_elements
```
jgebal: Really nice and clean test.

test/core/test_ut_utils.pkb
```diff
@@ -293,6 +293,26 @@ end;';
...
+    l_list_to_be_equal ut3.ut_varchar2_list := ut3.ut_varchar2_list('hello', 'world', 'okay');
+    l_list ut3.ut_varchar2_list := ut3.ut_varchar2_list(' hello  ', chr(9)||'world ', 'okay');
+  begin
+    --Act
+    l_list := ut3.ut_utils.trim_list_elements(l_list, '[:space:]');
+    --Assert
+    ut.expect(anydata.convertcollection(l_list)).to_equal(anydata.convertcollection(l_list_to_be_equal));
+  end;
 
+  procedure test_filter_list
```
jgebal: Can you add tests for NULL and empty collections?
Would be nice to know what will happen in those cases.

source/core/ut_suite_builder.pkb
```diff
@@ -41,12 +41,27 @@ create or replace package body ut_suite_builder is
...
       return a_index = a_annotations.count or a_annotations(a_index).subobject_name != nvl(a_annotations(a_index+1).subobject_name, ' ');
     end;
+
+    function exception_numbers_list(a_exception_numbers in varchar2) return ut_varchar2_list is
+      l_exception_number_list ut_varchar2_list;
+      l_regexp_for_excep_nums varchar2(30) := '^-?[[:digit:]]{1,5}$';
+    begin
+      /*the a_expected_error_codes is converted to a ut_varchar2_list after that is trimmed and filtered to left only valid exception numbers*/
+      l_exception_number_list := ut_utils.string_to_table(a_exception_numbers, ',', 'Y');
+      l_exception_number_list := ut_utils.trim_list_elements(l_exception_number_list, '[:space:]');
```
jgebal: can we have '[:space]' as default value for parameter?

source/core/ut_utils.pks
```diff
@@ -293,5 +293,11 @@ create or replace package ut_utils authid definer is
   */
   function to_xml_number_format(a_value number) return varchar2;
 
+  /*It takes a collection of type ut_varchar2_list and it trims the characters passed as arguments for every element*/
+  function trim_list_elements(a_list IN ut_varchar2_list, a_characters_to_trim in varchar2) return ut_varchar2_list;
```
jgebal: I would rename the parameter to be named `a_regexp_to_trim`, give it a default value of `[:space:]` or even consider removing the parameter as we don't have a use-case for trimming anything else than any blank characters at the moment.

test/core/annotations/test_annot_throws_exception.pkb
```diff
@@ -0,0 +1,187 @@
...
+    l_drop_statment VARCHAR2(32737);
+  begin
+    l_drop_statment := 'drop package annotated_package_with_throws';
+    execute immediate l_drop_statment;
+  end;
+
+  function execution_test_result(a_procedure_test_name in varchar) return varchar2 is
+    l_result varchar2(32737);
+  begin
+    select column_value
```
jgebal: Maybe run the suite once in the `beforeall`?
That way the tests will run much much faster.
Currently each individual test takes ~0.5 second to execute and the first one needs over 2 seconds (includes suite parsing for annotated_package_with_throws)
```
      annotations- throws
        Gives success when annotated number exception is thrown [2.606 sec]
        Gives succes when one of the annotated number exceptions is thrown [.474 sec]
        Gives failure when the raised exception is different that the annotated one [.457 sec]
        Ignores when  the annotation throws is empty [.488 sec]
        Ignores when only bad parameters are passed, the test raise a exception and it shows errored test [.524 sec]
        Ignores when only bad parameters are passed, the test does not raise a exception and it shows successful test [.491 sec]
        Detects a valid exception number within many invalid ones [.484 sec]
        Givess failure when a exception is expected and nothing is thrown [.529 sec]
```

source/core/types/ut_executable_test.tpb
```diff
@@ -0,0 +1,79 @@
...
+    l_completed_without_errors  boolean;
+  begin
+    l_completed_without_errors := self.do_execute(a_item, a_listener, a_expected_error_codes);
+  end do_execute;
+
+  member function do_execute(self in out nocopy ut_executable_test, a_item in out nocopy ut_suite_item, a_listener in out nocopy ut_event_listener_base,
+                              a_expected_error_codes in ut_varchar2_list := null) return boolean is
+    l_ut_executable ut_executable;
+    l_expected_except_message  varchar2(4000);
+    l_owner                    varchar2(200) := self.owner_name;
```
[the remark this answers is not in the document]
AlexisGaldamez: I did it that way because that way it's in *ut_test.do_execute*, can I refactor that in *ut_test.do_execute* too?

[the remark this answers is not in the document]
jgebal: Well, if it is not used as assignment in `ut_executable` then go ahead.
If there is room for improvement - improve it ;)

test/core/test_ut_utils.pkb
AlexisGaldamez: I will do it

source/core/ut_suite_builder.pkb
AlexisGaldamez: Yes, good idea

test/core/test_ut_utils.pkb
AlexisGaldamez: Thanks

source/core/ut_utils.pks
AlexisGaldamez: It's good if I add the default and change the parameter name. But I thinks I can be useful if you want to trim -?(),.;  something like that for a collection someday

source/core/types/ut_executable_test.tpb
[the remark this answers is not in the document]
AlexisGaldamez: Great :)

PR #565, opened 2018-01-29
Fix root tag for Sonar Unit Test reporter

rostskadat:
According to https://github.com/SonarQubeCommunity/sonar-generic-coverage#unit-tests-execution-results-report-format

### DIFF
--- a/source/reporters/ut_sonar_test_reporter.tpb
+++ b/source/reporters/ut_sonar_test_reporter.tpb
@@ -89,12 +89,12 @@ create or replace type body ut_sonar_test_reporter is
     end;
 
   begin
-    self.print_text('<testExecutions version="1">');
+    self.print_text('<unitTest version="1">');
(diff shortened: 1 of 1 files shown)


rostskadat:
This is a fix for #566

jgebal:
The element is correct, as sonar changed the root element.
The link you provided states in readme file.
```
Deprecated since SonarQube 6.2 because the functionality was absorbed into SonarQube core.
```

utPLSQL v3 was released way after sonar 6.2
We decided to support new version instead of the legacy one.

rostskadat:
OK

rostskadat:
We'll update our local instalation.

BTW, thanks for the amazing package :)

jgebal:
I would suggest creating a new reporter as a copy or better under this one.
That way you will not get impacted by changes done in future releases

PR #591, opened 2018-02-19
Add Diff functionality to cursors, collections and object types

### LINKED ISSUE #451: Add ability to extract cursor definition information
## Ability to extract cursor definition information
------------

When consructing `ut_data_value_refcursor`, we need to capture the cursor number using `dbm_sql.to_cursor_number` and store it as attribute of the object.
New type `ut_data_value_cursor_def` needs to be created to store description of cursor definition (column names and data types). 

Descriptor would be the attribute of `ut_data_value_refcursor`.

**Additionally** 
Extracting data from cursor into a table should be delayed and not performed when object is created but rather when the comparator is called.

### LINKED ISSUE #452: Add ability to compare cursor definition before comparing cursor data
## Ability to compare cursor definition before comparing cursor data
---------------------
The type `ut_data_value_cursor_def` should implement a comparator method `compare_implementaion` to allow comparison between descriptors.
The type needs to have `to_string` and a `diff` function, so that we can build a report that would contain:
- full descriptor of cursor
- diff of descriptors of tow cursors

Diff might be a challenge as it's hard to present a meaningful diff in a plain text format.

### LINKED ISSUE #560: Null attribute is reported differently on 11gR2 and 12c (r1/r2)
When reporting a mismatch on data between objects in 11g, null elements are missing (not shown at all).
On 12c however they are showing as `<attribute_name/>` - that is empty XML tag.
If we figure out how to resolve this, we should make it consistent across db versions.
Example:
```sql
create or replace type test_dummy_object as object (
  id number,
  "name"  varchar2(30),
  "Value" varchar2(30)
)
/
```

```sql
select xmltype.createXml( test_dummy_object(null, null, null) ) from dual ;
```
On 11g
```
<TEST_DUMMY_OBJECT/>
```

On 12c
```
<TEST_DUMMY_OBJECT><ID/>
<name/>
<Value/>
</TEST_DUMMY_OBJECT>
```

jgebal:
Reports generated on cursor/collection/object type will now show meaningful information about differences between actual and expected data.
Cursor comparison is now more strict - it will fail if cursor data-types are different.
Resolves #560
Resolves #452
Resolves #451


jgebal:
@john-otoole - will you be able to review documentation changes?

john-otoole:
Jacek, I've added some inline comments to commit
https://github.com/utPLSQL/utPLSQL/commit/6649502

jgebal:
Thanks @john-otoole I'll update the documentation accordingly.

PR #594, opened 2018-02-24
Feature/documentation throws annotation

### LINKED ISSUE #585: Add description of new `--%throws` annotation in documentation

jgebal:
Resolves #585
Updated documentation for `--%throws` annotation. 
Updated documentation to assure that annotations are mentioned as `--%annotation` rather than `%annotation`
Fixed bug, where `--%throws(-00001)` would not catch exception `ORA-00001` due to leading zero.
Added full stack trace, when a different than expected exception was thrown.
Added ut_integer_list_type.
Cleanup of `ut_executable_test`
Reworked tests for `--%throws` to check for the test report text content.


jgebal:
@AlexisGaldamez - can you please have a look at this?

PR #604, opened 2018-03-15
12.2 Oracle Native Block Coverage

### LINKED ISSUE #476: Add 12.2 coverage reporter
Extend existing coverage to support two new fields:
- branches_to_cover
- branches_covered

Values should be null for coverage using `dbms_profiler`

For coverage in 12.2 we need to convert block coverage into line/branches coverage to match generic test coverage data for Sonar.
https://docs.sonarqube.org/display/SONAR/Generic+Test+Data

htlm coverage reproter should be updated to support `yellow` color for branches not covered in 100% for line.

htlm coverage reproter should be updated to show `covered/to_cover` branches indicator in the report when there is more than 1 branch.

lwasylow:
Added a support for running cover coverage from Oracle Block Coverage.
Changes includes,update to html reporter for presentation of data,update to core types to support new coverage type option as well as few minor changes.
Default option remains profiler. The queries are in sys refcursors to avoid a compilation issues on older version of databases.
Fixes #476


source/core/coverage/ut_coverage.pkb
```diff
@@ -176,81 +102,18 @@ create or replace package body ut_coverage is
...
-
-    return l_result;
-  end get_coverage_data;
-
+    
+    if a_coverage_options.coverage_type = c_block_coverage then
+      $if dbms_db_version.version = 12 and dbms_db_version.release >= 2 or dbms_db_version.version > 12 $then
+       return ut_coverage_block.get_coverage_data_block(a_coverage_options => a_coverage_options);
+      $else
+       return null;
```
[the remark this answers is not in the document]
lwasylow: Its something we could consider. However if we want to warn then probably at the coverage_start function.
As dbms output ?

source/api/ut.pkb
```diff
@@ -106,29 +106,29 @@ create or replace package body ut is
 
   procedure run_autonomous(
     a_paths ut_varchar2_list, a_reporter ut_reporter_base, a_color_console integer,
-    a_coverage_schemes ut_varchar2_list := null, a_source_file_mappings ut_file_mappings, a_test_file_mappings ut_file_mappings,
+    a_coverage_schemes ut_varchar2_list := null,a_coverage_type varchar2:=null, a_source_file_mappings ut_file_mappings, a_test_file_mappings ut_file_mappings,
```
jgebal: The new parameter should be added as last to avoid breaking changes to API.

source/api/ut.pkb
```diff
@@ -106,29 +106,29 @@ create or replace package body ut is
...
+      ut_utils.int_to_boolean(a_color_console), a_coverage_schemes,a_coverage_type,
       a_source_file_mappings, a_test_file_mappings, a_include_objects, a_exclude_objects
     );
     rollback;
   end;
 
   procedure run_autonomous(
     a_paths ut_varchar2_list, a_reporter ut_reporter_base, a_color_console integer,
-    a_coverage_schemes ut_varchar2_list := null, a_source_files ut_varchar2_list, a_test_files ut_varchar2_list,
+    a_coverage_schemes ut_varchar2_list := null,a_coverage_type varchar2:=null, a_source_files ut_varchar2_list, a_test_files ut_varchar2_list,
```
jgebal: as above

source/core/coverage/ut_block_helper.pkb
```diff
@@ -0,0 +1,89 @@
+create or replace package body ut_block_helper is
```
jgebal: maybe name it `ut_block_coverage_helper` - easier to understand

source/core/coverage/ut_block_helper.pkb
```diff
@@ -0,0 +1,89 @@
...
+  */
+
+  type t_proftab_row is record (
+      line  binary_integer,
+      calls number(38,0)
+    );
+    
+  type t_proftab_rows is table of t_proftab_row;
+
+  type t_block_row is record(
```
jgebal: I would transform the not-so-good Oracle results into:

- start_line  binary_integer
- end_line  binary_integer - (`lead(line,1) over(order by line)` from coverage data - null if no next line)
- blocks     binary_integer
- covered_blocks binary_integer

I assume that there is no situation where we have more than 1 block spanning across more than one line.
That is:
only when start_line = end_line we can get blocks>1

source/core/coverage/ut_coverage_proftab.pkb
```diff
@@ -0,0 +1,178 @@
...
+  --   - PACKAGE, PROCEDURE, FUNCTION definition line,
+  --   - BEGIN, END  of a block
+  -- Another transformation is adjustment of line number for TRIGGER body.
+  -- DBMS_PROFILER is reporting line numbers for triggers not as defined in DBA_SOURCE, its usign line numbers as defined in DBA_TRIGGERS
+  -- the DBA_TRIGGERS does not contain the trigger specification lines, only lines that define the trigger body.
+  -- the query adjusts the line numbers for triggers by finding first occurrence of begin|declare|compound in the trigger body line.
+  -- The subquery is optimized by:
+  -- - COALESCE function -> it will execute only for TRIGGERS
+  -- - scalar subquery cache -> it will only execute once for one trigger source code.
+  function get_cov_sources_sql(a_coverage_options ut_coverage_options) return varchar2 is
```
jgebal: Isn't that function duplicated in two packages now?

source/core/coverage/ut_block_helper.pkb
Pazus: I’m afraid it might be possible for complex multi line `if` Statement

lwasylow: Even when simple if and bad beutifier we would have one block in single line.

source/core/coverage/ut_coverage_proftab.pkb
lwasylow: Not exactly, I  block coverage we do not skip any lines. I tried to keep to minimum number of places where there is a If decision. What I might do is to move it to generic helper and in profiler coverage package put some code to apply case statement for skipping lines before execution. So that main query will be maintained in one place.

source/core/coverage/ut_block_helper.pkb
lwasylow: Not sure if thats really do the trick as we cant really tell more details about column line of the block.Also graphical represenation on the html report or in fact even xml would struggle to present these data.

lwasylow:
Changes has been added for extended coverage and fixes implemented as per recommendation when required.

Issue #635, opened 2018-04-10
Reporter ut_xunit_reporter : Can't publish on jenkins


jgarec:
Hi,

Note sure if it's related with #634.
With Jenkins, there is something wrong with "Publish with xUnit test result report"

Running utplsql with command line: 
`utplsql run <user>/<password>@<db_host>:<port>/<SID> -f=ut_xunit_reporter -o=target/utplsql_xunit_reporter.xml [...]
`

Jenkins configuration : 
![image](https://user-images.githubusercontent.com/16031094/38562379-b6dad646-3cda-11e8-8fa2-f7cf3e02028b.png)

Log file :

> [xUnit] [INFO] - Starting to record.
> [xUnit] [INFO] - Processing JUnit
> [xUnit] [INFO] - [JUnit] - 1 test report file(s) were found with the pattern 'target/*.xml' relative to '/xxxxx/poc-utplsql' for the testing framework 'JUnit'.
> [xUnit] [ERROR] - The result file '/xxxxx/poc-utplsql/target/utplsql_xunit_reporter.xml' for the metric 'JUnit' is not valid. The result file has been skipped.
> [xUnit] [INFO] - Failing BUILD because 'set build failed if errors' option is activated.

Generated report file : 
[utplsql_xunit_reporter.txt](https://github.com/utPLSQL/utPLSQL/files/1895012/utplsql_xunit_reporter.txt)

Any idea ?
We are using a quite old version of jenkins (2.32).

lwasylow:
Please try using publisj junit results plugin instead.

jgarec:
I was not able to select "publish junit test results" because my project is a "Maven project" and plugin is hidded in this case.
https://issues.jenkins-ci.org/browse/JENKINS-9980

with a freestyle project it's ok.

lwasylow:
We will look to get proper junit format to work on Jenkins. Hopefully that resolve your issues.

PR #642, opened 2018-04-13
Fixes/xunit reporter

### LINKED ISSUE #634: UT_XUNIT_REPORTER format not conform https://gist.github.com/kuzuha/232902acab1344d6b578
In the ut_xunit_reporter spec it is specified that the XML generated is in a format conforming with JUnit4.
Looks like that's not the case. 
E.g. the attributes in node testsuites: failure, error and skipped should be resp. failures, errors and disabled.

lwasylow:
Fixes to xunit reporter.
current one is actually junit and is not conforming to junit4 xsd.
the synonym for xunit to junit has been created for backwards compatibility.
fixes #634 and #635


jgarec:
Hi,

Are you sure it's a good thing to keep this compatibility ?
I think It should be now a breaking change or it will be later when someone wants to create a real xunit reporter.

lwasylow:
Hi I believe so.
Since is quite popular as using jenkins and there was no real demand for xunit at the moment(?), better to have minimal impact on users where possible.

pesse:
I agree with @lwasylow to keep impact on existing setups as low as possible.
Alternative would be to not use a synonym but a wrapper-object which spits out a warning that xunit-reporter is "deprecated" in its current form. With 3.1.0 we have warnings.
What do you think, @jgebal?

lwasylow:
Its a bit like stuck between a rock and hard place.
None of it is ideal, as again deprecated suggest is no longer in use but in fact it wasn't in use ever as true xunit  :) 
Either way we will break someone build now or later and probably we will break it twice, once now and once when true xunit will be ready.

jgebal:
I'm not sure if we will ever introduce real XUnit. 
JUnit seems more than enough for integration with CI/CD.
I like the idea of depreciation.
We could add warning inXUnit  reporter constructor an the XUnit could be just a wrapper object built on top of JUnit reporter.
Public synonym would then point to XUnit reporter object.
Counts good?

lwasylow:
So to summarise, we will leave a xunit and xunit synonym but put a warning in constructor and from there we will call a new junit reporter that will be implemented as new reporter? Yeah sounds good.

old_tests/RunAll.sql
```diff
@@ -83,12 +83,12 @@ exec ut_coverage.coverage_start_develop();
 @@lib/RunTest.sql ut_reporters/ut_sonar_test_reporter.AcceptsFileMapping.sql
 @@lib/RunTest.sql ut_reporters/ut_sonar_test_reporter.ProducesExpectedOutputs.sql
 @@lib/RunTest.sql ut_reporters/ut_teamcity_reporter.ProducesExpectedOutputs.sql
-@@lib/RunTest.sql ut_reporters/ut_xunit_reporter.ProducesExpectedOutputs.sql
+@@lib/RunTest.sql ut_reporters/ut_junit_reporter.ProducesExpectedOutputs.sql
```
jgebal: Will you be able to migrate the `old_tests` that you modify to the new tests format (using utPLSQL)?

source/create_synonyms_and_grants_for_user.sql
```diff
@@ -127,7 +127,8 @@ create or replace synonym &ut3_user..match for &&ut3_owner..match;
 create or replace synonym &ut3_user..ut for &&ut3_owner..ut;
 create or replace synonym &ut3_user..ut_runner for &&ut3_owner..ut_runner;
 create or replace synonym &ut3_user..ut_teamcity_reporter for &&ut3_owner..ut_teamcity_reporter;
-create or replace synonym &ut3_user..ut_xunit_reporter for &&ut3_owner..ut_xunit_reporter;
+create or replace synonym &ut3_user..ut_xunit_reporter for &&ut3_owner..ut_junit_reporter;
```
jgebal: synonym should point to `ut_xunit_reporter`

source/reporters/ut_junit_reporter.tpb
```diff
@@ -35,9 +34,11 @@ create or replace type body ut_xunit_reporter is
     procedure print_test_elements(a_test ut_test) is
       l_lines ut_varchar2_list;
       l_output clob;
+      l_cdata_tag_start varchar2(10) := '<![CDATA[';
```
jgebal: make it constant?

source/reporters/ut_junit_reporter.tpb
```diff
@@ -46,16 +47,22 @@ create or replace type body ut_xunit_reporter is
       if a_test.result = ut_utils.gc_error then
         self.print_text('<error>');
         self.print_text('<![CDATA[');
```
jgebal: use the constant for tag name since it's already defined.

source/reporters/ut_junit_reporter.tpb
```diff
@@ -46,16 +47,22 @@ create or replace type body ut_xunit_reporter is
...
-        self.print_clob(ut_utils.table_to_clob(a_test.get_error_stack_traces()));
+        self.print_clob(
+        replace(replace(ut_utils.table_to_clob(a_test.get_error_stack_traces()),l_cdata_tag_start,''),l_cddata_tag_end,'')
```
jgebal: why do we need to do this replace?

source/reporters/ut_junit_reporter.tpb
```diff
@@ -78,10 +96,35 @@ create or replace type body ut_xunit_reporter is
...
+          print_suite_elements(treat(a_suite.items(i) as ut_logical_suite), a_suite_id);
+        end if;
+      end loop;
+      
+      -- Now when all testsuite are printed do the testcases.
+      for i in 1 .. l_tests.count loop
+       print_test_elements(treat(l_tests(i) as ut_test));
+      end loop;
+      
+      /*for i in 1 .. a_suite.items.count loop
```
jgebal: no need to keep commented code

source/reporters/ut_junit_reporter.tpb
```diff
@@ -99,36 +144,36 @@ create or replace type body ut_xunit_reporter is
...
-           || '" error="' || a_item.results_count.errored_count
-           || '" failure="' || a_item.results_count.failure_count
+    return ' disabled="' || a_item.results_count.disabled_count
+           || '" errors="' || a_item.results_count.errored_count
+           || '" failures="' || a_item.results_count.failure_count
            || '" name="' || dbms_xmlgen.convert(nvl(a_item.description, a_item.name))
            || '" time="' || ut_utils.to_xml_number_format(a_item.execution_time()) || '" ';
   end;
 
+  member function get_common_test_attributes(a_item ut_suite_item) return varchar2 is
```
jgebal: maybe name it `get_common_item_attributes` and also call it at line 167 - you will avoid duplicated lines for name/time

source/reporters/ut_junit_reporter.tpb
lwasylow: In rare occasions when cddata string is part of error message. It then breaks xml format.

old_tests/RunAll.sql
lwasylow: Sounds good

source/reporters/ut_xunit_reporter.tpb
```diff
@@ -23,115 +23,14 @@ create or replace type body ut_xunit_reporter is
   end;
 
   overriding member procedure after_calling_run(self in out nocopy ut_xunit_reporter, a_run in ut_run) is
```
jgebal: No need for override - in it's not overridden it will call the `after_calling_run` from parent type.

source/reporters/ut_xunit_reporter.tps
```diff
@@ -23,7 +23,6 @@ create or replace type ut_xunit_reporter under ut_output_reporter_base(
   constructor function ut_xunit_reporter(self in out nocopy ut_xunit_reporter) return self as result,
 
   overriding member procedure after_calling_run(self in out nocopy ut_xunit_reporter, a_run in ut_run),
```
jgebal: can be removed

source/uninstall.sql
```diff
@@ -54,6 +54,8 @@ drop synonym be_true;
 
 drop synonym equal;
 
+drop synonym ut_xunit_reporter;
```
jgebal: we don't have synonym anymore

PR #644, opened 2018-04-16
Feature/12ccoverage doc

lwasylow:
Updating Documentation regarding extended coverage


jgebal:
Looking at the images:
Where did the:
...lines missed. **tr> <** 
come from?
It's showing in the header line of `extended_coverage_html_summary.png`?

lwasylow:
Ah.. Trigger happy to push through so missed. I have taken out some things from html manually as first lines were at 0% in my sample i had. Will get that corrected and run tests on one test to get a nice result.

jgebal:
So nothing wrong with the code itself then?

lwasylow:
No it was just sample for whole ut3 schema so I removed some tr manually to make it more visible. I have rerun code for a one file now so it looks nicer on screenshot but code is OK.

PR #648, opened 2018-04-21
Feature/cursor implicit

### LINKED ISSUE #632: Cursors Comparision, lack of colum aliases results in XML parsing error.
Using 12.1 with sample test as follow:

```sql
CREATE OR REPLACE PACKAGE ut_sample_test IS

   --%suitepath(Root Tests)
   --%suite(Sample Test Suite)

   --%test(Compare Ref Cursors)
   PROCEDURE ut_refcursors;

END ut_sample_test;
/

CREATE OR REPLACE PACKAGE BODY ut_sample_test IS

   PROCEDURE ut_refcursors IS
      v_actual   SYS_REFCURSOR;
      v_expected SYS_REFCURSOR;
   BEGIN
      OPEN v_actual FOR
         SELECT 2,
                1 test
           FROM dual;
   
      OPEN v_expected FOR
         SELECT 1,
                1 test
           FROM dual;
   
      ut.expect(v_actual).to_equal(v_expected);
   END;

END ut_sample_test;
```

Executing test as follow:
```sql
SELECT *
  FROM TABLE(ut.run());
```
Results in Error 
``` sql
root tests
  Sample Test Suite
    Compare Ref Cursors [.01 sec] (FAILED - 1)
 
Failures:
 
  1) ut_refcursors
      ORA-31011: XML parsing failed
      ORA-19213: error occurred in XML processing at lines 1
      LPX-00231: invalid character 49 ('1') found in a Name or Nmtoken
      ORA-06512: at "UT3.UT_COMPOUND_DATA_HELPER", line 142
      ORA-06512: at "UT3.UT_DATA_VALUE_REFCURSOR", line 164
      ORA-06512: at "UT3.UT_EQUAL", line 217
      ORA-06512: at "UT3.UT_EXPECTATION", line 26
      ORA-06512: at "UT3.UT_EXPECTATION", line 136
      ORA-06512: at "SUPERDEV.UT_SAMPLE_TEST", line 15
      ORA-06512: at line 6
Finished in .011594 seconds
1 tests, 0 failed, 1 errored, 0 disabled, 0 warning(s)
 

```

lwasylow:
Implement fix for xml parsing when cursor column missing alias. Fixes #632 
This will not include changes to include and exclude options as this will require syntax checking with xpath and possible rework way options are passed

### DIFF
--- a/source/core/ut_utils.pkb
+++ b/source/core/ut_utils.pkb
@@ -544,5 +544,19 @@ procedure append_to_clob(a_src_clob in out nocopy clob, a_clob_table t_clob_tab,
     return l_filtered_list;
   end;
 
+  function xmlgen_escaped_string(a_string in varchar2) return varchar2 is
+    l_result varchar2(4000);
+    l_sql varchar2(32767) := q'!select q'[!'||a_string||q'!]' as "!'||a_string||'" from dual';
+  begin
+    if a_string is not null then
+      select extract(dbms_xmlgen.getxmltype(l_sql),'/*/*/*').getRootElement() 
+      into l_result
+      from dual;
+    else
+    l_result := a_string;
+    end if;
+    return l_result;
+  end;
+
 end ut_utils;
 /
--- a/source/core/ut_utils.pks
+++ b/source/core/ut_utils.pks
@@ -302,5 +302,8 @@ create or replace package ut_utils authid definer is
   /*It takes a collection of type ut_varchar2_list and it only returns the elements which meets the regular expression*/
   function filter_list(a_list IN ut_varchar2_list, a_regexp_filter in varchar2) return ut_varchar2_list;
 
+  -- Generates XMLGEN escaped string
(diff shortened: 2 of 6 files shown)


jgebal:
Just to be clear.
With this change, we allow for comparison of columns with implicit names, however, we do not support implicit column names in the include/exclude filters.
Is that right?

lwasylow:
That's correct, I have tested filters and problem is that there is no clear distinct between a column name in filter or xpath expression. E. G. ('Test', '\\/test') without knowing these we would ass that second value is a special name for column where in fact was invalid xpath.

jgebal:
Can you add a note to `advanced_data_comparison.md` to inform that exclude/include does not support columns with implicit names?
We could also add a test to prove that it's not working - as an expected behavior.

The test should document, what happens today if we use the `exclude()` with implicit column name?

lwasylow:
Of course :+1:

PR #666, opened 2018-05-11
Feature/cursor pk join

### LINKED ISSUE #453: Add ability to join cursors by specific columns(PK/UK)
## Ability to join cursors by specific columns(PK/UK)
---------------------------

This feature could/should be implemented in a similar fashion as excluding columns with `a_exclude` parameter.

We would have additional parameter for the `ut_equal` matcher if the matcher is used with refcursor.

If the parameter is present, we can transform it into XPATH to extract PK/UK/join specific column values from the data and store them in a separate VARCHAR2 column of the temp table for cursor data (with limit of 4000 bytes).
There are few ways of extracting column data from XML - XSL
https://stackoverflow.com/questions/17314062/how-to-convert-xml-to-csv-using-xsl
or 
Extract with XPath:
```
select extract(dbms_xmlgen.getxmltype('
       select object_name, object_type
       from user_objects
       '),'//OBJECT_NAME/text()|//OBJECT_TYPE/text()') xml
from dual;
```

The temp table keys column could be indexed with a unique index, so that we can actually validate that the join columns are unique.

If columns are not unique, the expectation should fail with meaningful message informing that the cursor data contains duplicates. 
We need to be able to inform the user which of the compared cursors doesn't have uniqueness which is again tricky as the data_value are not aware of where they were created from (actual or expected).

lwasylow:
Changes to implement join by cursor and unordered compare.
fixes #453


jgebal:
@lwasylow , can you update the branch and resolve conflicts

test/core/expectations/compound_data/test_expectations_cursor.pkb
```diff
@@ -1041,5 +1041,405 @@ Rows: [ 2 differences ]%
...
+    l_expected SYS_REFCURSOR;
+  begin
+    --Arrange
+    open l_actual for select owner, object_name,object_type from all_objects where owner = user
+    order by 1,2,3 asc;
+    open l_expected for select owner, object_name,object_type from all_objects where owner = user
+    order by 1,2,3 desc;
+    
+    --Act
+    ut3.ut.expect(l_actual).to_equal(l_expected).join_by('OWNER');
```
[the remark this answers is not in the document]
lwasylow: We are partition row number over PK HASH and join on pk_hash and duplicate no.
Because duplicate_no in both sets match we got success.
If we have an extra row we would get extra row error. 
But agree that test is probably not the best example of join on uniquness.

test/core/expectations/compound_data/test_expectations_cursor.pkb
```diff
@@ -1041,5 +1041,405 @@ Rows: [ 2 differences ]%
...
+    open l_actual   for select rownum as rn, 'x' SOME_COL  from dual a connect by level < 4;
+    open l_expected for select rownum as rni, 'x' SOME_COL  from dual a connect by level < 4;
+     
+    --Act
+    ut3.ut.expect(l_actual).to_equal(l_expected).join_by('RNI');
+    --Assert
+ l_expected_message := q'[%Actual: refcursor [ count = 3 ] was expected to equal: refcursor [ count = 3 ]%
+Diff:%
+%Unable to join sets:%
+%Unknown key to join by in actual:/*/RNI%]';
```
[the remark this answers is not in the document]
lwasylow: Challenge here is that we do not know what column name is missing only what XPATH expression to join by is not present. Because we allow user to pass in Xpath expression.
The message  
```sql
Expected: <USER_ID>-600</USER_ID> for key: TEST
```
Is extracted from piece of XML using xpath.

In scenario where there is missing key we cannot use that XPATH to extract information about column and we cannot assume that xpath expression is a column name and not some xpath search string with wildcard.

test/core/expectations/compound_data/test_expectations_cursor.pkb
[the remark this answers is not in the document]
jgebal: Right, so in fact we are preventing cartesian joins and if key is not unique we would do a join by a rownum within join key value.
Worth documenting - I already forgot it was supposed to work like this

test/core/expectations/compound_data/test_expectations_cursor.pkb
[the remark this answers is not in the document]
jgebal: Can you do a testcase for join on nested table element in cursor column?
Would be interesting to see such join happening.

test/core/expectations/compound_data/test_expectations_cursor.pkb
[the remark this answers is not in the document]
lwasylow: Documentation has been updated with more details

test/core/expectations/compound_data/test_expectations_cursor.pkb
```diff
@@ -1041,5 +1041,405 @@ Rows: [ 2 differences ]%
...
+Diff:%
+%Unable to join sets:%
+%Unknown key to join by in expected:/*/SOME_COL%
+%Unknown key to join by in actual:/*/SOME_COL%]';
+    l_actual_message := ut3.ut_expectation_processor.get_failed_expectations()(1).message;
+    --Assert
+    ut.expect(l_actual_message).to_be_like(l_expected_message);
+  end;
+  
+  procedure cur_joinby_comp_twocols_exkey is
```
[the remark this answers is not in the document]
lwasylow: Updated docs

test/core/expectations/compound_data/test_expectations_cursor.pkb
```diff
@@ -1041,5 +1041,405 @@ Rows: [ 2 differences ]%
...
+    open l_actual   for select rownum as rn, 'x' SOME_COL  from dual a connect by level < 4;
+    open l_expected for select rownum as rn, 'x' SOME_COL  from dual a connect by level < 4;
+     
+    --Act
+    ut3.ut.expect(l_actual).to_equal(l_expected).join_by('OWNER');
+    --Assert
+ l_expected_message := q'[%Actual: refcursor [ count = 3 ] was expected to equal: refcursor [ count = 3 ]%
+Diff:%
+%Unable to join sets:%
+%Unknown key to join by in expected:/*/OWNER%
```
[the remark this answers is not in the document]
lwasylow: Updated to refactor new name

test/core/expectations/compound_data/test_expectations_cursor.pkb
```diff
@@ -1041,5 +1041,516 @@ Rows: [ 2 differences ]%
...
+    open l_actual for select username, user_id from all_users union all
+    select 'TEST' username, -610 user_id from dual order by 1 asc;
+     
+    --Act
+    ut3.ut.expect(l_actual).to_equal(l_expected).join_by(ut3.ut_varchar2_list('USERNAME,USER_ID'));
+    --Assert
+ l_expected_message := q'[%Actual: refcursor [ count = % ] was expected to equal: refcursor [ count = % ]
+%Diff:%
+%Rows: [ 2 differences ]%
+%Missing:  <USERNAME>TEST</USERNAME><USER_ID>-600</USER_ID> for key: -600:TEST%
```
jgebal: I have a doubt about readability of this message to the user.
` Missing:  <USERNAME>TEST</USERNAME><USER_ID>-600</USER_ID> for key: -600:TEST`
If we don't join by PK, we use syntax like
`  Row No. 3 - Missing:  <RN>3</RN>`
How about if we would present it to the user in a similar way:
`PK <USERNAME>TEST</USERNAME><USER_ID>-600</USER_ID> missing.`

jgebal: I think we should exclude PK columns from the Missing/Extra data list - this is duplicated info.

test/core/expectations/compound_data/test_expectations_cursor.pkb
```diff
@@ -1041,5 +1041,516 @@ Rows: [ 2 differences ]%
...
+    open l_actual for select username, user_id,'Y' is_valid from all_users union all
+    select 'TEST' username, -610 user_id,'Y' is_valid from dual order by 1 asc;
+     
+    --Act
+    ut3.ut.expect(l_actual).to_equal(l_expected).join_by(ut3.ut_varchar2_list('USERNAME,IS_VALID'));
+    --Assert
+ l_expected_message := q'[%Actual: refcursor [ count = % ] was expected to equal: refcursor [ count = % ]
+%Diff:%
+%Rows: [ 1 differences ]%
+%Expected: <USER_ID>-600</USER_ID> for key: TEST:Y%
```
jgebal: I believe, this could be more readable if it was:
`PK <USERNAME>TEST</USERNAME><IS_VALID>Y</IS_VALID> - Expected: <USER_ID>-600</USER_ID>`
`PK <USERNAME>TEST</USERNAME><IS_VALID>Y</IS_VALID> - Actual:   <USER_ID>-610</USER_ID>`

test/core/expectations/compound_data/test_expectations_cursor.pkb
```diff
@@ -1041,5 +1041,516 @@ Rows: [ 2 differences ]%
...
+    open l_actual for select test_dummy_object( rownum, 'Something '||rownum, rownum) as colval
+      from dual connect by level <=2 order by rownum asc;
+
+    open l_expected for select test_dummy_object( rownum, 'Somethings '||rownum, rownum) as colval
+      from dual connect by level <=3 order by rownum desc;
+    
+    --Act
+    ut3.ut.expect(l_actual).to_equal(l_expected).unordered;
+    --Assert
+    ut.expect(expectations.failed_expectations_data()).to_be_not_null();
```
jgebal: What data will the user see in this case?
Can you test for an expected message?

test/core/expectations/compound_data/test_expectations_cursor.pkb
```diff
@@ -1041,5 +1041,516 @@ Rows: [ 2 differences ]%
...
+    open l_actual   for select rownum as rn, 'x' SOME_COL  from dual a connect by level < 4;
+    open l_expected for select rownum as rn, 'x' SOME_COL  from dual a connect by level < 4;
+     
+    --Act
+    ut3.ut.expect(l_actual).to_equal(l_expected).join_by(ut3.ut_varchar2_list('RN,SOME_COL')).exclude('RN');
+    --Assert
+ l_expected_message := q'[%Actual: refcursor [ count = 3 ] was expected to equal: refcursor [ count = 3 ]%
+Diff:%
+%Unable to join sets:%
+%Join key /*/RN does not exists in expected%
```
jgebal: Can we remove the leading `/*/` from the message?
That way we get actual column name rather than XPATH.
Not sure if it makes sense but probably 98% of use-cases will be joins by regular columns (not nested elements)

test/core/expectations/compound_data/test_expectations_cursor.pkb
```diff
@@ -1041,5 +1041,516 @@ Rows: [ 2 differences ]%
...
+    --Arrange
+    open l_actual   for select rownum as rn, 'TEST' as A_COLUMN  from dual a connect by level < 4;
+    open l_expected for select rownum as rn, 1 as A_COLUMN  from dual a connect by level < 4;
+    --Act
+    ut3.ut.expect(l_actual).to_equal(l_expected).include(ut3.ut_varchar2_list('RN')).exclude(ut3.ut_varchar2_list('A_COLUMN'));
+    --Assert
+    ut.expect(expectations.failed_expectations_data()).to_be_empty();
+  end;
+    
+  procedure compare_nested_tab_col_un is
```
jgebal: The test is actually for an object in a cursor column, not a nested table.
Can you rename those tests and add an additional test for a nested table (like `ut_varchar2_list` or so)?
I wonder how it will work with a nested table containing more than one row.
Maybe we should not support that at all?

test/core/expectations/compound_data/test_expectations_cursor.pkb
lwasylow: Agree, will update

lwasylow: Make sense

test/core/expectations/compound_data/test_expectations_cursor.pkb
lwasylow: Make sense

test/core/expectations/compound_data/test_expectations_cursor.pkb
lwasylow: Yeah we can make that assumption just have to updated docs to mention that.

test/core/expectations/compound_data/test_expectations_cursor.pkb
lwasylow: that looks fine select * from table(ut_varchar2_list) for one or two columns ut_key_value_pairs

test/core/expectations/compound_data/test_expectations_cursor.pkb
```diff
@@ -1540,17 +1551,150 @@ Diff:%
...
+      
+    --Arrange
+    open l_actual for select key,value
+      from table(l_actual_tab) order by 1 asc;
+
+    open l_expected for select key,value
+      from table(l_expected_tab) order by 1 desc;
+    
+    --Act
+    ut3.ut.expect(l_actual).to_equal(l_expected).join_by('KEY');
```
jgebal: What will happen when your cursor `l_actual` is:
```sql
select rownum rn, l_actual_tab
  from dual connect by level <=2;
ut3.ut.expect(l_actual).to_equal(l_expected).join_by('RN,L_ACTUAL_TAB/KEY');
```
The dataset is then something like:
```xml
<?xml version="1.0"?>
<ROWSET>
 <ROW>
  <RN>1</RN>
  <L_ACTUAL_TAB>
   <UT_KEY_VALUE_PAIR>
    <KEY>1</KEY>
    <VALUE>Something 1</VALUE>
   </UT_KEY_VALUE_PAIR>
   <UT_KEY_VALUE_PAIR>
    <KEY>2</KEY>
    <VALUE>Something 2</VALUE>
   </UT_KEY_VALUE_PAIR>
  </L_ACTUAL_TAB>
 </ROW>
 <ROW>
  <RN>2</RN>
  <L_ACTUAL_TAB>
   <UT_KEY_VALUE_PAIR>
    <KEY>1</KEY>
    <VALUE>Something 1</VALUE>
   </UT_KEY_VALUE_PAIR>
   <UT_KEY_VALUE_PAIR>
    <KEY>2</KEY>
    <VALUE>Something 2</VALUE>
   </UT_KEY_VALUE_PAIR>
  </L_ACTUAL_TAB>
 </ROW>
</ROWSET>
```
That will not be (and probably should not be) supporter but might be still worth documenting in test.
We just don't know what will happen as there is not test case to document how the `ut.expect` will behave.

lwasylow:
Sure thing

PR #670, opened 2018-05-13
Ability to run test even when package specs of test are invalid

### LINKED ISSUE #489: Allow for execution of test with invalid package spec.
Currently, if a unit test package specification gets invalidated due to a broken dependency or dependency recreation, the unit test package is skipped in execution.

This is something to be fixed, as only package spec that has errors cannot be executed.
utPLSQL should try to parse annotations of invalid packages and try to execute them.
If a package fails to be validated when invoked, the test would fail.

lwasylow:
Ability to run test even when package specs of test are invalid due to rebuild of dependent package or broken dependency.
It will try to parse and execute test package which can result in two situation:
1) package got validated by call and test continue
2) package is still invalid and test fail with a message that test procedure is invalidated.
Fixes #489


source/core/types/ut_executable.tps
```diff
@@ -1,4 +1,4 @@
-create or replace type ut_executable under ut_event_item(
+create or replace type ut_executable force under ut_event_item(
```
jgebal: why change to force?

test/api/test_ut_run.pkb
```diff
@@ -114,5 +114,107 @@ Failures:%
...
+  end;
+
+  procedure run_and_revalidate_specs is
+    l_results   ut3.ut_varchar2_list;
+    l_actual    clob;
+    l_is_invalid number;
+  begin
+    execute immediate q'[select count(1) from all_objects o where o.owner = :object_owner and o.object_type = 'PACKAGE'
+            and o.status = 'INVALID' and o.object_name= :object_name]' into l_is_invalid
+            using user,'FAILING_INVALID_SPEC';
```
jgebal: It was tricky to understand what test is doing because the name used in this test `FAILING_INVALID_SPEC` is quite misleading.
Maybe change the name to something more like: `invalid_package_that_revalidates`?

lwasylow: done

PR #677, opened 2018-05-22
 Allow before/after annotations to accept list of [[owner.]package.]procedure

### LINKED ISSUE #649: Allow before/after annotations to accept list of [[owner.]package.]procedure
I'd like to suggest the following to the `before...`, `after...` annotations.
- add ability to reference procedures from other packages
- add ability to call multiple procedures

Additionally, `beforeall`, `afterall`, `beforeeach`, `aftereach`, we could be suite level annotations with similar syntax as `beforetest`/`aftertest`.
Example below:
```sql
package test_departments_api is

  --%suite(Departments API)
  --%beforeall(common_data_helper.data_setup, privileges_helper.setup)
  --%afterall(privileges_helper.cleanup, common_data_helper.data_cleanup)


  procedure setup_manager;

  --%test(Adds department)
  --%beforetest(some_package.add_employee, some_package.add_location, setup_manager)
  procedure successfully_add_department;

  --%test(Fails to add depatrment when no manager found)
  --%beforetest(common_employee_helper.setup_employee, common_location_helper.setup_location)
  procedure fail_to_add_department;

  --%test(Returns department by name)
  --%beforetest(common_department_helper.setup_depatment)
  procedure get_department_by_name;

end;
```

This way, we can allow for even better separation of common things.
It is still possible to achieve the same without `before`/`after` annotations referencing procedures outside of package, but it requires a lot of bolierplate code as below.

```sql
package test_departments_api is

  --%suite(Departments API)

  --%beforeall
  procedure common_setup;
  
  --%afterall
  procedure common_cleanup;

  procedure setup_manager;

  --%test(Adds department)
  --%beforetest(some_package.add_employee, some_package.add_location, setup_manager)
  procedure successfully_add_department;

  --%test(Fails to add depatrment when no manager found)
  --%beforetest(common_employee_helper.setup_employee, common_location_helper.setup_location)
  procedure fail_to_add_department;

  --%test(Returns department by name)
  --%beforetest(common_department_helper.setup_depatment)
  procedure get_department_by_name;

end;
/
...
package body test_departments_api is

  procedure common_setup is
  begin
    common_data_helper.data_setup;
    privileges_helper.setup;
  end;

  procedure common_cleanup is
  begin
    privileges_helper.cleanup;
    common_data_helper.data_cleanup;
  end;

...
```
Would it make sence and add value to have this kind of flexibility in annotations?

The down side I see is that we would only know at **runtime** if the dependencies to other packages defined by before/after are valid.

AlexisGaldamez:
ut_suit_builder modified to accept a list of procedures external o internal to the test package.

Resolve #649

### DIFF
--- a/test/core/annotations/test_before_after_test_annotation.pkb
+++ b/test/core/annotations/test_before_after_test_annotation.pkb
@@ -1,0 +1,138 @@
+create or replace package body test_before_after_annotations is
+  g_tests_results clob;
+
+  procedure create_tests_results is
+    pragma autonomous_transaction;
+
+    l_test_package_spec  varchar2(32737);
+    l_test_package_body  varchar2(32737);
+    l_dummy_utility_pkg_spec varchar2(32737);
+    l_dummy_utility_pkg_body varchar2(32737);
+    l_test_results  ut3.ut_varchar2_list;
+    
+    procedure drop_package(a_package_name varchar2)
+    is
+    begin
+      execute immediate 'drop package '||a_package_name;
+    end;
+  begin
+    l_test_package_spec := '
+        create or replace package dummy_before_after_test is
+            --%suite(Package to test annotations beforetest and aftertest)
+            
+            l_dummy_1 integer;
+            l_dummy_2 integer;
+            l_dummy_3 integer;
+            
+            --%beforeeach
+            procedure clean_global_variables;
+
+            --%test(Beforetest with call to procedure external to the test package)
+            --%beforetest(dummy_utility_pkg_1.test1)
+            procedure beforetest_one_ext_procedure;
+            
+            
+            --%test(Beforetest with call to multi procedures external and interal to the test package)
+            --%beforetest(dummy_utility_pkg_1.test1, test2, ut3_tester.dummy_utility_pkg_1.test3)
+            procedure beforetest_multi_ext_procedure;  
+            
+            --%test(Beforetest with call to multi procedures where one does not exist)
+            --%beforetest(dummy_utility_pkg_1.test1, non_existent_procedure, ut3_tester.dummy_utility_pkg_1.test3)
+            procedure beforetest_one_err_procedure;
+            
+            procedure test2;
+        end;
+    ';
+
+    l_test_package_body := '
+        create or replace package body dummy_before_after_test is
+            procedure clean_global_variables is
+            begin
+              l_dummy_1 := null;
+              l_dummy_2 := null;
+              l_dummy_3 := null;  
+            end;
+        
+            procedure beforetest_one_ext_procedure is
+            begin
+              ut3.ut.expect(l_dummy_1).to_equal(1);
+            end;
+            
+            procedure beforetest_multi_ext_procedure
+            is
+            begin
+              ut3.ut.expect(l_dummy_1).to_equal(1);  
+              ut3.ut.expect(l_dummy_2).to_equal(2);
+              ut3.ut.expect(l_dummy_3).to_equal(3);
+            end;
+            
+            procedure beforetest_one_err_procedure
+            is
+            begin
+              null;
+            end;
+            
+            procedure test2 is
+            begin
+              l_dummy_2 := 2;  
+            end;
+        end;
+    ';
+    
+    l_dummy_utility_pkg_spec := '
+      create or replace package dummy_utility_pkg_1 is
+        procedure test1;
+        
+        procedure test3;
+      end;
+    ';
+    
+    l_dummy_utility_pkg_body := '
+      create or replace package body dummy_utility_pkg_1 is
+        procedure test1 is
+        begin
+          dummy_before_after_test.l_dummy_1 := 1;
+        end;
+        
+        procedure test3 is
+        begin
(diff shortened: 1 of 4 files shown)


test/core/annotations/test_before_after_test_annotation.pks
```diff
@@ -0,0 +1,18 @@
...
+
+  --%test(Beforetest with call to procedure external to the test package)
+  procedure beforetest_one_ext_procedure;
+  
+  --%test(Beforetest with call to multi procedures external and interal to the test package)
+  procedure beforetest_multi_ext_procedure; 
+  
+  --%test(Beforetest with call to multi procedure where one does not exist)
+  procedure beforetest_one_err_procedure;
+end;
```
jgebal: There is a `/` missing - the test package is not compiling from `sqlcl`

test/core/annotations/test_before_after_test_annotation.pkb
```diff
@@ -0,0 +1,138 @@
...
+    is
+    begin
+      execute immediate 'drop package '||a_package_name;
+    end;
+  begin
+    l_test_package_spec := '
+        create or replace package dummy_before_after_test is
+            --%suite(Package to test annotations beforetest and aftertest)
+            
+            l_dummy_1 integer;
```
jgebal: I would put those variables into `dummy_utility_pkg_1`.
That way we avoid bidirectional dependency where test package depending on utility and vice versa.

test/core/annotations/test_before_after_test_annotation.pkb
```diff
@@ -0,0 +1,138 @@
...
+  begin
+    ut.expect(g_tests_results).to_match('^\s*Beforetest with call to multi procedures external and interal to the test package \[[\.0-9]+ sec\]\s*$','m');
+    ut.expect(g_tests_results).not_to_match('beforetest_multi_ext_procedure');
+  end;
+  
+  procedure beforetest_one_err_procedure
+  is
+  begin
+    ut.expect(g_tests_results).not_to_match('^\s*Beforetest with call to multi procedures where one does not exist \[[\.0-9]+ sec\]\s*$','m');
+    ut.expect(g_tests_results).to_match('beforetest_one_err_procedure');
```
jgebal: Can you make the expectation on full failure message?
That way it's easier to understand what is the expected behavior.

test/core/annotations/test_before_after_test_annotation.pkb
```diff
@@ -0,0 +1,138 @@
...
+  is
+  begin
+    ut.expect(g_tests_results).to_match('^\s*Beforetest with call to multi procedures external and interal to the test package \[[\.0-9]+ sec\]\s*$','m');
+    ut.expect(g_tests_results).not_to_match('beforetest_multi_ext_procedure');
+  end;
+  
+  procedure beforetest_one_err_procedure
+  is
+  begin
+    ut.expect(g_tests_results).not_to_match('^\s*Beforetest with call to multi procedures where one does not exist \[[\.0-9]+ sec\]\s*$','m');
```
jgebal: Can you make expectation so that it says what `is` expected rather than what `is not` expected?
I *guess*, we expect a failing test here, but the line isn't explicit about that.

test/core/annotations/test_before_after_test_annotation.pkb
```diff
@@ -0,0 +1,138 @@
...
+  is
+  begin
+    ut.expect(g_tests_results).to_match('^\s*Beforetest with call to procedure external to the test package \[[\.0-9]+ sec\]\s*$','m');
+    ut.expect(g_tests_results).not_to_match('beforetest_one_ext_procedure');
+  end;
+  
+  procedure beforetest_multi_ext_procedure
+  is
+  begin
+    ut.expect(g_tests_results).to_match('^\s*Beforetest with call to multi procedures external and interal to the test package \[[\.0-9]+ sec\]\s*$','m');
```
jgebal: We should probably check that:
- each of the `beforetest` was actually executed
- they were executed in the order they are specified

test/core/annotations/test_before_after_test_annotation.pkb
jgebal: What will happen with the third `beforetest` procedure when the second one cannot be found?
Will it be executed or not?
It would be great to be able to check that.
Maybe you could have a global boolean variable in `dummy_utility_pkg_1` for each procedure to check if procedure was executed and then have expectation on that?
```sql
  v_test1 boolean;
  v_test2 boolean;
  v_test3 boolean;
  v_beforetest_one_ext boolean;
  v_beforetest_multi_ext boolean;
  v_beforetest_one_err boolean;
```

AlexisGaldamez: Actually, when a beforetest is not found it does not execute the others procedures either the test. For example:
```
  --%beforetest(dummy_utility_pkg_1.test4, non_existent_procedure, ut3_tester.dummy_utility_pkg_1.test4)
  procedure beforetest_one_err_procedure;
```
It only will execute dummy_utility_pkg_1.test4 nothing more. is is the behaviour we want to?

PR #678, opened 2018-05-23
Test to verify there's a problem with cursor-comparison on 12.1

pesse:
when there are several numbers with different precision and we make use of package-level nested table types


jgebal:
I wasn't able to reproduce this at work on 12.1
Seems that one of patches for Oracle DB is fixing this issue.
Our Docker databases are bare images with no patches appied wich can be the reason for this error.
Seems to me like an Oracle bug that was already fixed, but ee can't test fir it on our instances.

jgebal:
@pesse 
Do you have any ideas on documentaing this isssue other than by PR?
Maybe a dosabled test or an issue?

pesse:
What do you think about a "Known Issues" document? Because we will obviously not fix this thing because you should just patch your Oracle.
Also: Should we try updating our docker images to a patched Oracle version?

jgebal:
Looks like this is no longer an issue.
The test was added to prevent regression in future.

PR #720, opened 2018-07-16
Fixed issue with sonar scanner failing on PR's from other repositories.

jgebal:


jgebal:
I think you're right.

Issue #734, opened 2018-07-24
Create build for Oracle Database 18c


Pazus:
Configure new item in build matrix to test utPLSQL against 18.3 version of the database

PR #735, opened 2018-07-25
Tests for Oracle Database 18c

Pazus:


jgebal:
Can you update documentation/readme where we mention DB versions supported?

jgebal:
Resolves #734

PR #738, opened 2018-07-27
Updated sqlcl version used for tests to version 18.2

Pazus:
Updated sqlcl version used for tests to version 18.2


jgebal:
Great work @Pazus

PR #750, opened 2018-08-11
Resolves issues with teamcity reporter

### LINKED ISSUE #747: TeamCity reporter not properly working with cursor comparison
**Background:**

Doing a cursor comparison like this one:
```
    PROCEDURE PRC_T_SAMPLE_4
    IS
        l_expected   v_ref_cursor;
        l_actual     v_ref_cursor;
    BEGIN
        OPEN l_expected FOR
            SELECT * FROM SAMPLE_TABLE WHERE filter=2; --about 20 columns and 50 rows

        OPEN l_actual FOR
            SELECT * FROM SAMPLE_TABLE WHERE filter=3; --about 20 columns and 50 rows

        ut.expect (l_expected).to_equal (l_actual);
    END;
```
If the test fail on the teamcity report we will have something like: 
##teamcity[testFailed timestamp='2018-08-03T10:11:14.221+0200' details='Actual: |r ....... ' Name='name' ]

Inside the details the xml structure of the actual cursor and the expected cursor can be found. 

**Issue**

Teamcity is no able to detect the testFailed. So teamcity report it as green (success) when it failed, it shall be red.

**Posible problem**

I opened the build log of team city and the line with all XML structure was infact in diferent lines with line breaks (every 4000 characters).  So i think the problem is that the output line of the Teamcity reporter was to big and teamcity was not able to parse it.

**Posible fix**

Trim the details string to 2500 or 3000 characters so the ##teamcity is less than 4000.

jgebal:
- the `[` and `]` were not escaped
- the `"` was escaped though not needed
- If total length of message inside the square brackets exceeds 4000 characters it's causing teamcity to report success for a test failed
Resolves #747

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -2,14 +2,14 @@
 # http://www.mkdocs.org/user-guide/configuration/
 
 site_name: utPLSQL
-site_description: utPLSQL Documenation Powerful Unit Testing Framework for Oracle PL/SQL
+site_description: utPLSQL Ultimate Unit Testing Framework for Oracle PL/SQL
 copyright: Copyright &copy; 2016 - 2018 utPLSQL Team
 repo_url: https://github.com/utPLSQL/utPLSQL
 theme: mkdocs
 use_directory_urls: false
 strict: true
 
-pages:
+nav:
   - Home: index.md
   - User Guide:
        - Installation: userguide/install.md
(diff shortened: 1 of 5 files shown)


jgebal:
Need to add one more test fir long outputs. If output is long I cut it and the message can be misleadimg as it doesn't say it's trimmed.

PR #836, opened 2019-02-20
Create CODE_OF_CONDUCT.md

jgebal:


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,76 @@
...
+representing a project or community include using an official project e-mail
+address, posting via an official social media account, or acting as an appointed
+representative at an online or offline event. Representation of a project may be
+further defined and clarified by project maintainers.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported by contacting the project team at utplsql@utplsql.org. All
+complaints will be reviewed and investigated and will result in a response that
```
pesse: Who is getting this e-mail? What do we suggest if a complaint is about one the "core maintainers"? Should we list different possibilities or encourage to reach out to someone they trust via Slack?

jgebal: At the moment I'm receiving the emails from that email box.
Though I must admit that I don't feel comfortable being solely responsible for it.
You're right. Slack would be better (for me anyway)

pesse: Just to be clear: This is not about mistrust - there are view people I would trust more to handle any complaints than you to be honest - it's about giving possibilities and future proof.
I'd suggest the slack-group (which is auto-invite and therefore entirely open) and list a number of individuals in that section:
Jacek, Pavel, me, Philipp, Lukasz (not sure if it makes sense to list more)

PR #849, opened 2019-03-06
Fix/sonar analysis shell

lwasylow:
TRAVIS_PULL_REPO_SLUG is empty when there is no PR.
The change introduced earlier added AND condition which was only met when there was actual PR.
Changed logic to look at the BRANCH if its not PR or PR from same REPO SLUG


lwasylow:
This is a test PR for now to test sonar pr

PR #878, opened 2019-03-26
Added support for random order of test execution.

### LINKED ISSUE #422: Random order of test execution
Define suite level annotation to shuffle tests on execution

jgebal:
Resolves #422


jgebal:
**Documentation update is needed**
We need to include description of the random order parameters.

lwasylow:
There is also fix for #914

PR #888, opened 2019-03-31
Feature/two output buffers

jgebal:
This feature will require changes on cli.
See: utPLSQL/utPLSQL-cli#140


PhilippSalvisberg:
I installed the branch `feature/two_output_buffers` in my local environment, run the tests and got the following result:

```
Finished in 79.545574 seconds
797 tests, 0 failed, 0 errored, 1 disabled, 0 warning(s)
```

I will need to adapt the current implementation for the SQL Developer extension on my branch `issue_6` since I'm using this incompatible code:

```
SELECT t.item_type, t.text
  FROM table(ut_output_table_buffer(?).get_lines()) t
```

I'll change that to the way suggested in https://github.com/utPLSQL/utPLSQL-cli/issues/140 - That is not a problem at all, especially since it is a development branch only.

I see this PR as a quick fix for the problems introduced with the change from VC2 to CLOB. However, I think we can do better in terms of API design (making it more narrow, less redundant and more independent of the implementation) without compromising performance. This can be addressed in dedicated issues and coming releases.

PR #892, opened 2019-04-02
Bugfix/issue 891

PhilippSalvisberg:
used `to_match` instead of `to_be_like` to make it work on all 11.2.0.4 Oracle Database instances.

### DIFF
--- a/docs/userguide/expectations.md
+++ b/docs/userguide/expectations.md
@@ -272,7 +272,9 @@ begin
 end;
 ```
 
-Parameters `a_mask` and `a_escape_char` represent valid parameters of the [Oracle LIKE condition](https://docs.oracle.com/database/121/SQLRF/conditions007.htm#SQLRF52142)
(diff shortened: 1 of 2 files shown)


lwasylow:
Would not be better to create a new test called it fix_to_bug_891 and make a changes there ? 
That way we kept original test behavior but also created and working test/example of workaround for possible issues ? 
(and user if required can disable a test in develop ? )

PhilippSalvisberg:
@lwasylow the behavior of the test did not change. It just uses `to_match` instead of `to_be_like`. The tests test exactly the same thing. It's a simple solution with minimal impact.

PhilippSalvisberg:
Please not that the bug kicks in only when comparing CLOB with LIKE using ESCAPE, and even than not in all cases. To avoid that we simply have to use REGEXP_LIKE instead of LIKE with ESCAPE. in this case this means `match` instead of `be_like`. It's not really a fix. It's a documented workaround.

PhilippSalvisberg:
Currently we are just lucky to use a 11.2 XE docker image, where this bug does not exist. This solution ensures that it will work even on the most recent 11.2 version. From my point of view this is better than having some tests which need to be manually enabled/disabled.

lwasylow:
Make senses.

PR #893, opened 2019-04-02
Feature/introduce testing with accurate grants

### LINKED ISSUE #867: Move all/most of tests to minimal privileged user
To avoid bugs related to privileges sneaking through in our self-testing we should do the following:

- use user `ut3_tester` to invoke all the tests - as we do today
- move all the tests to `ut3$user#` - user without elevated privileges

This way we can assure that users without `execute any procedure` privilege can use all of our framework without issues.

lwasylow:
Fixes #867
It introduces a new user `ut3$user#` that have minimum privileges and move all public API tests into that schema. User `ut3_tester` is a internal framework testing user who tests functions not exposed to public. 
Last user is one that calls CLI and is elevated privileges user `ut3_tester_helper` this user is created for purpose of allowing setup and teardown some test data for other users as well as provide some methods used during expectations check.


development/cleanup.sh
```diff
@@ -22,16 +21,17 @@ end;
 drop user ${UT3_OWNER} cascade;
 drop user ${UT3_RELEASE_VERSION_SCHEMA} cascade;
 drop user ${UT3_TESTER} cascade;
+drop user ${UT3_TESTER_HELPER} cascade;
 drop user ${UT3_USER} cascade;
 
 begin
   for i in (
-    select decode(owner,'PUBLIC','drop public synonym "','drop synonym "'||owner||'"."')|| synonym_name ||'"' drop_orphaned_synonym, owner||'.'||synonym_name syn from dba_synonyms a
```
jgebal: Can we keep that functionality?

development/template.env.sh
```diff
@@ -13,7 +13,10 @@ export UT3_OWNER_PASSWORD=ut3
 export UT3_RELEASE_VERSION_SCHEMA=ut3_latest_release
 export UT3_TESTER=ut3_tester
 export UT3_TESTER_PASSWORD=ut3
+export UT3_TESTER_HELPER=ut3_tester_helper
+export UT3_TESTER_HELPER_PASSWORD=ut3
 export UT3_TABLESPACE=users
 export UT3_USER="UT3\$USER#"
 export UT3_USER_PASSWORD=ut3
-
+export UT3_TESTER_HELPER=ut3_tester_helper
```
jgebal: Duplicate with line 16

source/uninstall_objects.sql
```diff
@@ -137,10 +137,6 @@ drop type ut_data_value_boolean force;
 
 drop type ut_data_value_blob force;
 
-drop type ut_data_value_object force;
```
jgebal: Why those drops are removed?

lwasylow: This are the types that been removed with introduction of ANYDATA into cursor. I can create separate issue on this .

development/cleanup.sh
lwasylow: Hmm good catch I dont remember changing this. I will double check whats going on.

lwasylow: I have branched out of random order , must some changes there that were not pulled back.

development/template.env.sh
lwasylow: Removed duplicates

source/uninstall_objects.sql
jgebal: Cool

jgebal:
Only 6 tests are missing now - Develop has 799 tests: https://travis-ci.org/utPLSQL/utPLSQL/jobs/514642945#L6244

lwasylow:
This are the minimum test user which are no longer required as the whole public api tests are minimal user

lwasylow:
This are the minimum test user which are no longer required as the whole public api tests are minimal user

jgebal:
I'm doing a tiny change - will push it soon and we can go ahead with the PR

PR #906, opened 2019-04-24
Feature/tags

### LINKED ISSUE #66: Tags for tests and suites
Introduce annotations to tag tests and suites.
Implement filters on execution procedures to include/exclude suites/tests by tags.

We should discuss if "tag" should be assigned to a suite or "suite_type" with predetermined values fits bettet.

lwasylow:
Fixes #66 
Implements a tags as a new argument a_tags that accepts a coma separated list of tags.
Tags are placed on suite and test level.
The tags are filter applied after suitepath is resolved.
The logic of multiple tags is `OR` so at least one tag must match for test/suite to be included.


source/core/ut_suite_manager.pkb
```diff
@@ -376,12 +383,16 @@ create or replace package body ut_suite_manager is
...
       l_path := lower( a_path );
     end if;
-
-    open l_result for
+    l_suite_item_name := case when l_tag_list is not null then 'suite_items_tags' else 'suite_items' end;
+    
+    
+    
+    /* Rewrite that as tags should be put on whats left not on full suite item cache */
+    l_sql :=
```
jgebal: It might be good to have a chat on that query. I'd like to understand how it works after changes.

source/core/ut_suite_manager.pkb
```diff
@@ -659,10 +692,39 @@ create or replace package body ut_suite_manager is
...
+  begin
+    if a_suite_path is not null then 
+      l_error_msg := 'No suite packages found for path '||a_schema_name||':'||a_suite_path;
+    elsif a_procedure_name is not null then 
+      l_error_msg := 'Suite test '||a_schema_name||'.'||a_object_name|| '.'||a_procedure_name||' does not exists';
+    elsif a_object_name is not null then
+      l_error_msg := 'Suite package '||a_schema_name||'.'||a_object_name|| ' does not exists';
+    end if;
+    
+    if l_error_msg is null and a_tags is not null then
```
jgebal: I'm not sure if we should raise if run is empty due to `tag` filters`.
If we execute `ut.run()` we get no errors of there are no suites in schema.
Maybe we should ignore non-existing tags and only filter by those that actually exist?

source/core/ut_suite_manager.pkb
```diff
@@ -677,7 +739,11 @@ create or replace package body ut_suite_manager is
     resolve_schema_names(l_paths);
 
     l_schema_paths := group_paths_by_schema(l_paths);
-
+    
+    -- TODO : use a_tags to filter out path. Should we do it before ? we dont know suites, if we do it after
```
jgebal: Needs to be discussed?

source/core/ut_suite_manager.pkb
```diff
@@ -192,7 +194,7 @@ create or replace package body ut_suite_manager is
                   results_count => ut_results_counter(), transaction_invalidators => ut_varchar2_list(),
                   items => a_items_at_level(a_prev_level),
                   before_all_list => sort_by_seq_no( a_rows( a_idx ).before_all_list), after_all_list => sort_by_seq_no(
-                    a_rows( a_idx ).after_all_list)
+                    a_rows( a_idx ).after_all_list), tags => a_rows(a_idx).tags --TODO : Should be share or separate
```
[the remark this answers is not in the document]
lwasylow: It was a notes when I was writing a code.
Should remove them.

source/core/ut_suite_manager.pkb
lwasylow: That was a note before we have decided the suitepath will be a primary filter and tags will filter remaining code.

source/core/ut_suite_builder.pkb
```diff
@@ -583,6 +621,31 @@ create or replace package body ut_suite_builder is
...
+    a_tags_ann_text tt_annotation_texts
+  ) is
+    l_annotation_pos binary_integer;
+    l_tag_list ut_varchar2_list := ut_varchar2_list();
+  begin
+    l_annotation_pos := a_tags_ann_text.first;
+    while l_annotation_pos is not null loop
+      if a_tags_ann_text(l_annotation_pos) is null then
+        a_suite.put_warning(
+            '"--%tag" annotation requires a tag value populated. Annotation ignored, line ' || l_annotation_pos
```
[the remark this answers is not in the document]
lwasylow: done

source/core/ut_suite_builder.pkb
```diff
@@ -305,6 +307,36 @@ create or replace package body ut_suite_builder is
...
+    a_tags_ann_text tt_annotation_texts
+  ) is
+    l_annotation_pos binary_integer;
+    l_tag_list ut_varchar2_list := ut_varchar2_list();
+  begin
+    l_annotation_pos := a_tags_ann_text.first;
+    while l_annotation_pos is not null loop
+      if a_tags_ann_text(l_annotation_pos) is null then
+        a_suite.put_warning(
+            '"--%tag" annotation requires a tag value populated. Annotation ignored.'
```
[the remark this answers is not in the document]
lwasylow: done

source/core/ut_suite_builder.pkb
```diff
@@ -583,6 +621,31 @@ create or replace package body ut_suite_builder is
...
+            '"--%tag" annotation requires a tag value populated. Annotation ignored, line ' || l_annotation_pos
+        );
+      else
+        l_tag_list := l_tag_list multiset union distinct ut_utils.trim_list_elements(
+          ut_utils.string_to_table(a_tags_ann_text(l_annotation_pos),',')
+          );
+      end if;
+      l_annotation_pos := a_tags_ann_text.next(l_annotation_pos);
+    end loop;
+    a_suite.tags := ut_utils.table_to_clob(
```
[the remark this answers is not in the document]
lwasylow: It removes empty elements from the table e.g. in annotation --%tags(tag1,,tag2) 
we got three tags: tag1,tag2 and empty string. 
I have used existing function to filter list element.

source/core/ut_suite_builder.pkb
```diff
@@ -305,6 +307,36 @@ create or replace package body ut_suite_builder is
...
+    l_annotation_pos := a_tags_ann_text.first;
+    while l_annotation_pos is not null loop
+      if a_tags_ann_text(l_annotation_pos) is null then
+        a_suite.put_warning(
+            '"--%tag" annotation requires a tag value populated. Annotation ignored.'
+            || chr( 10 ) || 'at "' || get_qualified_object_name(a_suite, a_procedure_name) || '", line ' || l_annotation_pos
+        );
+      else
+        l_tag_list := l_tag_list multiset union distinct ut_utils.trim_list_elements(
+          ut_utils.string_to_table(a_tags_ann_text(l_annotation_pos),',')
```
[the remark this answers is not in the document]
lwasylow: de-duplication is done using mulitset union distinct.
Not sure what you mean by storage limit ? Tag list is a temporary table list that that will get change into a flat varchar2(4000) using table to string.

source/core/ut_suite_builder.pkb
```diff
@@ -594,6 +657,7 @@ create or replace package body ut_suite_builder is
     l_after_all_list     tt_executables;
     l_rollback_type      ut_utils.t_rollback_type;
     l_annotation_text    t_annotation_text;
+    l_suite_tags         ut_varchar2_rows;
```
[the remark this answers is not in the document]
lwasylow: Its not. Cleared

source/core/ut_suite_manager.pkb
```diff
@@ -360,13 +363,17 @@ create or replace package body ut_suite_manager is
...
-    a_random_seed      positive
+    a_random_seed      positive,
+    a_tags             varchar2 := null
   ) return t_cached_suites_cursor is
     l_path     varchar2( 4000 );
     l_result   sys_refcursor;
     l_ut_owner varchar2(250) := ut_utils.ut_owner;
+    l_sql      varchar2(32767);
+    l_suite_item_name varchar2(20);
+    l_tag_list varchar2(4000) := a_tags;
```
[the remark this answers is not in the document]
lwasylow: Cleared. Left from when a_tags was a varchar2_rows

source/core/ut_suite_manager.pkb
lwasylow: This error is when we are calling a ut.run with a test that doesn't exists. This causing a suite to not be found. 
I think we should do same we do for suitepath e.g. raise an error for non existing suitepath as the tag is an explicit filter on tests we want to execute.
If we dont find any the user should be aware of that.

source/core/ut_suite_cache_tag.sql
```diff
@@ -0,0 +1,7 @@
+create table ut_suite_cache_tag (
```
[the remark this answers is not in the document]
lwasylow: I will test performance against nested table of ut_varchar2_rows and report back :)

source/core/types/ut_suite.tps
```diff
@@ -26,8 +26,13 @@ create or replace type ut_suite  under ut_logical_suite (
   * Procedure exists within the package of the suite
   */
   after_all_list ut_executables,
+  /**
+  * Hold list of tags assign to suite
+  */
+  tags varchar2(4000),
```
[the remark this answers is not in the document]
lwasylow: That makes sense. I will do that

docs/userguide/annotations.md
```diff
@@ -1221,6 +1222,91 @@ Finished in .035261 seconds
 ```
 
 
+
+### Tags
+
+Tag is a label attached to the test or a suite path. It is used for identification and execution a group of tests / suites that share same tag.  
+
+It allows us to group a tests / suites using a various categorization and place a test / suite in multiple buckets. Same tests can be group with other tests based on the functionality , frequency, type of output etc.
```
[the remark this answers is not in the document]
lwasylow: done

docs/userguide/annotations.md
```diff
@@ -1221,6 +1222,91 @@ Finished in .035261 seconds
...
+    open v_actual   for select 2 as test from dual;
+
+      ut.expect(v_actual).to_equal(v_expected);
+   END;
+   
+END ut_sample_test;
+/
+```
+
+Execution of the test is done by using a new parameter `a_tags`
```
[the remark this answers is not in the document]
lwasylow: good point :)

source/core/ut_suite_manager.pkb
lwasylow: sure

docs/userguide/annotations.md
```diff
@@ -1221,6 +1222,98 @@ Finished in .035261 seconds
...
+   PROCEDURE ut_refcursors1;
+
+   --%test(Run equality test)
+   --%tag(test2,sample)
+   PROCEDURE ut_test;
+   
+END ut_sample_test;
+/
+
+create or replace PACKAGE BODY ut_sample_test IS
```
jgebal: We promote all lower-case code in our examples.

source/core/ut_suite_manager.pkb
```diff
@@ -689,16 +747,13 @@ create or replace package body ut_suite_manager is
...
           );
         if a_suites.count = l_suites_count then
-          if l_path_item.suite_path is not null then
-            raise_application_error(ut_utils.gc_suite_package_not_found,'No suite packages found for path '||l_schema||':'||l_path_item.suite_path|| '.');
-          elsif l_path_item.procedure_name is not null then
-            raise_application_error(ut_utils.gc_suite_package_not_found,'Suite test '||l_schema||'.'||l_path_item.object_name|| '.'||l_path_item.procedure_name||' does not exist');
-          elsif l_path_item.object_name is not null then
-            raise_application_error(ut_utils.gc_suite_package_not_found,'Suite package '||l_schema||'.'||l_path_item.object_name|| ' does not exist');
-          end if;
+          raise_application_error(ut_utils.gc_suite_package_not_found,build_no_test_error_msg(
```
jgebal: Before the change, this was not throwing exception if running on a whole schema (with `ut.run()`) and schema has no tests.
This will be a breaking change in behavior as empty schema, schema without tests will start throwing exceptions.
Seems like we are missing a test to document existing behavior.

source/core/ut_suite_manager.pkb
```diff
@@ -659,10 +687,40 @@ create or replace package body ut_suite_manager is
...
+  begin
+    if a_suite_path is not null then 
+      l_error_msg := 'No suite packages found for path '||a_schema_name||':'||a_suite_path;
+    elsif a_procedure_name is not null then 
+      l_error_msg := 'Suite test '||a_schema_name||'.'||a_object_name|| '.'||a_procedure_name||' does not exists';
+    elsif a_object_name is not null then
+      l_error_msg := 'Suite package '||a_schema_name||'.'||a_object_name|| ' does not exists';
+    end if;
+    
+    if l_error_msg is null and a_tags.count > 0 then
```
jgebal: I would refrain from raising an error in this case.
If we pass 3 paths, and 1 tak and one of paths is not tagged, I would expect utPLSQL to execute the tagged items from remaining two paths.

If given suite doesn't have any of given tags it is OK, we should not throw exception, just run 0 (zero) tests.

source/core/ut_suite_builder.pkb
```diff
@@ -583,6 +625,30 @@ create or replace package body ut_suite_builder is
     a_suite.path := lower(coalesce(a_suite.path, a_suite.object_name));
   end;
 
+  procedure add_tags_to_suite(
```
jgebal: Seems like duplicate with `add_tags_to_test`
Any way we can make that function more generic?
Maybe we should have `add_tags_to_suite_item`?

source/api/ut_suite_item_info.tps
```diff
@@ -22,6 +22,10 @@ create or replace type ut_suite_item_info as object (
   item_type        varchar2( 250 ), -- the type of item (UT_SUITE/UT_SUITE_CONTEXT/UT_TEST)
   item_line_no     integer,         -- line_number where annotation identifying the item exists
   path             varchar2( 4000 ),-- suitepath of the item
-  disabled_flag    integer          -- 0 (zero) if item is not disabled, 1 if item is disabled by --%disabled annotation
+  disabled_flag    integer,          -- 0 (zero) if item is not disabled, 1 if item is disabled by --%disabled annotation
+  tags             varchar2(4000),
```
jgebal: maybe we should make it CLOB, then there is no size limit?

source/core/ut_suite_cache.sql
```diff
@@ -45,7 +46,8 @@ create table ut_suite_cache (
   nested table after_each_list store as ut_suite_cache_after_each
   nested table before_test_list store as ut_suite_cache_before_test
   nested table after_test_list store as ut_suite_cache_after_test
-  nested table expected_error_codes store as ut_suite_cache_trhows
+  nested table expected_error_codes store as ut_suite_cache_throws
+  nested table tags store as ut_suite_cache_tags return as locator
```
jgebal: Why `as locator`? Any notable performance gains?
If so, maybe we should consider the same for the remaining nested tables.

source/core/ut_suite_manager.pkb
```diff
@@ -17,6 +17,7 @@ create or replace package body ut_suite_manager is
   */
 
   gc_suitpath_error_message constant varchar2(100) := 'Suitepath exceeds 1000 CHAR on: ';
+  gc_tag_errmsg             constant number := 450;
```
jgebal: Sonar complains about `NUMBER` type without precision - change this to `INTEGER`.

source/core/ut_suite_manager.pkb
jgebal: The query I've created is quite a monster itself.
Do you think it would be beneficial if it is changed to be a template, like we did in cursor_compare?

test/ut3_tester/core/test_suite_builder.pkb
```diff
@@ -1114,5 +1133,305 @@ create or replace package body test_suite_builder is
...
+    ut.expect(l_actual).to_be_like(
+        '%<UT_LOGICAL_SUITE>' ||
+        '%<NAME>some_package</NAME><DESCRIPTION>testsuite</DESCRIPTION><PATH>some_package</PATH>' ||
+        '%<TAGS><VARCHAR2>suitetag</VARCHAR2><VARCHAR2>suitetag1</VARCHAR2><VARCHAR2>suitetag2</VARCHAR2></TAGS>%'||
+        '%</UT_LOGICAL_SUITE>%'
+    );
+ 
+  end;
+
+  procedure test_empty_tag_between is
```
jgebal: What about tag that has spaces inside?
`--%tags( my first tag, my second tag, some other tag )`
How will that work?
Are spaces in tags allowed?
Is comma the only disallowed character?

test/ut3_user/api/test_ut_run.pkb
```diff
@@ -787,5 +787,214 @@ Failures:%
...
+    --Assert
+    ut.expect( l_results ).to_be_like( '%test_package_1%' );
+    ut.expect( l_results ).to_be_like( '%test_package_2%' );
+    ut.expect( l_results ).to_be_like( '%test_package_3%' ); 
+  end; 
+  
+  procedure two_test_run_by_two_tags is
+    l_results clob;
+  begin
+    ut3_tester_helper.run_helper.run(a_tags => 'subtest1,subtest2');
```
jgebal: Can this list accept spaces in between?
What will happen if I pass `a_tags => 'subtest1, , subtest2'`? Will it still work?

test/ut3_user/api/test_ut_run.pkb
```diff
@@ -787,5 +787,214 @@ Failures:%
...
+  begin
+    ut3_tester_helper.run_helper.run(a_tags => 'suite2');
+    l_results :=  ut3_tester_helper.main_helper.get_dbms_output_as_clob();
+    --Assert
+    ut.expect( l_results ).to_be_like( '%test_package_1%' );
+    ut.expect( l_results ).to_be_like( '%test_package_2%' );
+    ut.expect( l_results ).not_to_be_like( '%test_package_3%' );  
+  end;
+  
+  procedure test_nonexists_tag is
```
jgebal: This is unexpected behavior.
We should change this.
When running all tests and filtering by tags, and no tagged items found, the run returns 0 (zero) tests executed.

test/ut3_user/api/test_ut_run.pkb
```diff
@@ -787,5 +787,214 @@ Failures:%
...
+  begin
+    ut3_tester_helper.run_helper.run(a_tags => 'suite1,suite1');
+    l_results :=  ut3_tester_helper.main_helper.get_dbms_output_as_clob();
+    --Assert
+    ut.expect( l_results ).to_be_like( '%test_package_1%' );
+    ut.expect( l_results ).to_be_like( '%test_package_2%' );
+    ut.expect( l_results ).not_to_be_like( '%test_package_3%' ); 
+  end;  
+ 
+  procedure run_proc_pkg_name_no_tag is
```
jgebal: Non existing tags...
There are two ways to go about it:
- ignore them
- filter everything
I would go with filter/reject everything.
But still, we shouldn't raise on "no suites found for tag"

test/ut3_user/api/test_ut_runner.pkb
```diff
@@ -247,12 +257,57 @@ end;';
...
-             'dummy_test_package.some_dummy_test_procedure' path, 0 disabled_flag
+             'dummy_test_package.some_dummy_test_procedure' path, 0 disabled_flag,null tags
+        from dual;
+    --Act
+    open l_actual for select * from table(ut3.ut_runner.get_suites_info('UT3$USER#','DUMMY_TEST_PACKAGE'));
+    --Assert
+    ut.expect(l_actual).to_equal(l_expected);
+  end;
+
+  --TODO : Apostro?????
```
jgebal: What is the TODO item here?

source/core/ut_suite_manager.pkb
lwasylow: Definitely I will do that

test/ut3_tester/core/test_suite_builder.pkb
lwasylow: The spaces are not allowed it will ignore such tag.

test/ut3_user/api/test_ut_run.pkb
lwasylow: It will filter out any empty spaces tags like one above, it will also trim them and remove any tags that have a space between words.

test/ut3_user/api/test_ut_runner.pkb
lwasylow: Note to myself :) to be removed

test/ut3_user/api/test_ut_run.pkb
lwasylow: done

source/api/ut_suite_item_info.tps
lwasylow: how it will look in the query ? the idea iof API is to be user friendly clob is not really. 
And 4k for tag is a really low case scenario

source/core/ut_suite_builder.pkb
lwasylow: Yes no that we moved tags into suite_items it makes sense

source/core/ut_suite_cache.sql
lwasylow: So I think depends on use in query we use it makes sense as per [docs](https://docs.oracle.com/en/database/oracle/oracle-database/18/adobj/design-considerations-for-collections.html#GUID-1CCB041E-AE30-4732-B39B-695E3291DC57) since we access for every row we look for a child record with a tag.
`For large child sets, the parent row and a locator to the child set can be returned so that the child rows can be accessed on demand; the child sets also can be filtered. Using nested table locators enables you to avoid unnecessarily transporting child rows for every parent. `

Having run locally tests with a just console reporter for locator and non locator run times been comparable , I will try do some autotrace when got some time.

PR #923, opened 2019-05-12
Anydata join / exclude / include

### LINKED ISSUE #912: exclude / join_by on collections
When using exclude / include / join_by with collections, we are forced to use syntax:
'object_type_name/attribute' for those features.

As far as I remember it used to work differently and you only needed to use attribute name.

lwasylow:
Fixes #912
Address issue of using a full type name in include / exclude / join_by on collections compare.
Now we are using only attributes names as previously.

We are treating refcursor as extra wrapping layer so when filtering a columns we need to make sure that a root name of collection / object name type is added to the filter to allow as properly use filter on cursor information.


source/core/ut_utils.pkb
```diff
@@ -798,5 +798,20 @@ create or replace package body ut_utils is
     return l_valid_name;
   end;
 
+  function add_prefix(a_list ut_varchar2_list, a_prefix varchar2, a_connector varchar2 := '/') return ut_varchar2_list is
```
jgebal: How about having add_prefix function for individual elements to avoid duplicating the below logic in 3 places:
```sql
        case 
          when self.is_anydata = 1 then
            l_prefix||'/'||trim (leading '/' from fl.column_value)
          else
            fl.column_value
          end
```

jgebal:
It would be great to enhance documentation for extended data comparison with 1-2 examples for exclude/include with use of ANYDATA.convertCollection.

source/expectations/data_values/ut_cursor_details.tpb
```diff
@@ -146,13 +146,25 @@ create or replace type body ut_cursor_details as
 
   member function get_missing_join_by_columns( a_expected_columns ut_varchar2_list ) return ut_varchar2_list is
     l_result ut_varchar2_list;
+    l_prefix            varchar2(125);
   begin
+      if self.is_anydata = 1 then
+        l_prefix := get_root;
```
jgebal: I would rather call it `l_root` everywhere.
It took me a while to understand what is the meaning of `l_prefix`.

source/core/ut_utils.pkb
lwasylow: good catch

pesse:
Never used it so far - maybe I should.
What's the benefit compared to cursor comparison? I remember to have tried comparing collections based on rowtypes and had problems with it...

lwasylow:
> 
> 
> Never used it so far - maybe I should.
> What's the benefit compared to cursor comparison? I remember to have tried comparing collections based on rowtypes and had problems with it...

You don't have to wrap a collection into cursor and then use a fancy include / exclude option to compare elements or object. An extra abstraction layer, underneath is still a cursor compare but accessing a nested elements directly to avoid an wrapping into long XML path.

source/core/ut_utils.pkb
lwasylow: Actually that causes an issue due to SQL context and min grant user.
We will need to create a public synonym to utils

lwasylow:
That's working however using a complex string manipulation. To simplify method the new attribute of filter path can be added but a diff sql will have to be changed to use a cursor details instead of xml extract

PR #926, opened 2019-05-23
Failure details in sonar test reporter are now CDATA

### LINKED ISSUE #925: Sonar test results fail on import.
**Describe the bug**
`ut_sonar_test_reporter` can produce invalid output on failing tests with long failure message.

The error that we see in sonar-scanner is:
```
Look at the SonarQube documentation to know the expected XML format. [com.ctc.wstx.exc.WstxLazyException] Unexpected character (CTRL-CHAR, code 10); expected a semi-colon after the reference for entity 'lt'
[ERROR] at [row,col {unknown-source}]: [4670,4001]
```

**Provide version info**
Running on latest utPLSQL v3.1.6

Not relevant

**To Reproduce**
Have a test that is failing and produces more than 4000 chars of output data and be unlucky enough to get the `&lt;` string at 4000 chars so the reporter inserts a line break.

Example:
```
   ... &lt;INSERT_TMSTMP&gt;2018-07-28T01:01:37.092127&lt;/INSERT_TMSTMP&gt;&lt
;GROUP_ID/&gt;...
```

**Expected behavior**
We should probably make the failure message as CDATA rather than escape the chars.

### LINKED ISSUE #927: JUnit test results import fail
**Describe the bug**
`ut_junit_reporter` can produce invalid output on failing tests with long failure message.

The error that we see in import of test results in Jenkins is:
```
Failed to read test report file [...snap...]/test_results.xml
org.dom4j.DocumentException: Error on line 6527 of document  : The reference to entity "lt" must end with the ';' delimiter. Nested exception: The reference to entity "lt" must end with the ';' delimiter.
```

**Provide version info**
Running on latest utPLSQL v3.1.6

Not relevant

**To Reproduce**
Have a test that is failing and produces more than 4000 chars of output data and be unlucky enough to get the `&lt;` string at 4000 chars so the reporter inserts a line break.

Example:
```
   ... &lt;INSERT_TMSTMP&gt;2018-07-28T01:01:37.092127&lt;/INSERT_TMSTMP&gt;&lt
;GROUP_ID/&gt;...
```

**Expected behavior**
We should probably make the failure message as CDATA rather than escape the chars.

jgebal:
Resolves #925 
Resolves #927


jgebal:
@lwasylow I've replaced the test results failures with escaped CDATA blocks.
This is to avid the issue I've seen at work (quite rare but still annoying), where if failure message is longer than 4k an dbms_xmlgen.convert is used, the line can get split into two lines inside the escaped XML characters making the XML report impossible to parse.

Let me know if you're OK with this change.

jgebal:
Seems like sonar is complaining on code duplication.
Maybe we should have some base XML functions to avoid duplication?

lwasylow:
Yeah definitely change make sense. And I think moving some of duplicate code into utils type is a good idea. Less maintenance and possibility of bugs.

PR #931, opened 2019-06-01
Fixes #921 - Assert input params in dynamic SQL

### LINKED ISSUE #921: Assert all input parameters used in dynamic SQL and PL/SQL
### Description

This is a generic issue. There are some places in utPLSQL that are vulnerable to SQL injections. We have 

1) find a way to identify them 
2) fix them
3) set up an automated review mechanism to find such issues, e.g. during our automated builds

@krisrice pointed out the following example in [ut_suite_manager.get_cached_suite_data](https://github.com/utPLSQL/utPLSQL/blob/develop/source/core/ut_suite_manager.pkb#L394)

```
and c.object_owner = ']'||upper(a_object_owner)||q'['
```

This line as two issues 

1) the parameter is not asserted, in this case via `sys.dbms_assert.schema_name`
2) the parameter is converted to upper case, which leads to wrong results when quoted usernames are used (see also #920).

### utPLSQL Version

v3.1.7.2808-develop

PhilippSalvisberg:
Used dbms_assert.qualified_sql_name whenever possible.
This is much less invasive than dbms_assert.enquote_name.
Code converting case is left untouched.

### DIFF
--- a/source/core/annotations/ut_annotation_manager.pkb
+++ b/source/core/annotations/ut_annotation_manager.pkb
@@ -57,7 +57,15 @@ create or replace package body ut_annotation_manager as
     l_objects_view varchar2(200) := ut_metadata.get_objects_view_name();
     l_cursor_text  varchar2(32767);
     l_result       ut_annotation_objs_cache_info;
+    l_object_owner varchar2(250);
+    l_object_type  varchar2(250);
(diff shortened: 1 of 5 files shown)


jgebal:
Resolves #921

PR #939, opened 2019-06-15
Removed dependency on `dbms_utility.name_resolve`.

### LINKED ISSUE #569: get rid of calls to dbms_utility.name_resolve if possible
Calls to `dbms_utility.name_resolve` are causing issues in the environments where schema_name and an object within the schema share the same name.
This was raised as an issue in utPLSQL itself:
https://stackoverflow.com/questions/45350935/utplsql-installation-ora-04047-error-when-exec-ut-run/48541758#48541758
And here is a bit of background to that issue from one of IBM forums:
https://www.ibm.com/developerworks/community/forums/html/topic?id=d8087c9d-39c2-4cf3-8e29-8611f48a4d6e

Need to review the need to call  `dbms_utility.name_resolve` functionality as well as ways to efficiently replace it (if it's really needed)

### LINKED ISSUE #885: "Call params for test are not valid" for all tests in specific schema
**Describe the bug**
The error message happens for all tests in a specific schema

**Provide version info**
utPLSQL v3.1.6.2735, Database version 18.5.0.0.0
port_string: x86_64/Linux 2.4.xx

- SQL Developer
- CLI

**To Reproduce**
1. Create betwnstr/test_betweenstr packages from demo project
2. EXECUTE ut.run('SCHEMA.test_betwnstr');
3. See error

Between string function
  Returns substring from start position to end position [,003 sec] (FAILED - 1)
  Returns substring when start position is zero [,002 sec] (FAILED - 2)
  Returns string until end if end position is greater than string length [,002 sec] (FAILED - 3)
  Returns null for null input string value [,002 sec] (FAILED - 4)
  A demo of test raising runtime exception [,002 sec] (FAILED - 5)
  A demo of failing test [,002 sec] (FAILED - 6)
  Demo of a disabled test [0 sec] (DISABLED)
 
Failures:
 
  1) normal_case
      Call params for test are not valid: package does not exist or is invalid: SCHEMA.TEST_BETWNSTR
  2) zero_start_position
      Call params for test are not valid: package does not exist or is invalid: SCHEMA.TEST_BETWNSTR
  3) big_end_position
      Call params for test are not valid: package does not exist or is invalid: SCHEMA.TEST_BETWNSTR
  4) null_string
      Call params for test are not valid: package does not exist or is invalid: SCHEMA.TEST_BETWNSTR
  5) bad_params
      Call params for test are not valid: package does not exist or is invalid: SCHEMA.TEST_BETWNSTR
  6) bad_test
      Call params for test are not valid: package does not exist or is invalid: SCHEMA.TEST_BETWNSTR
Finished in ,014138 seconds
7 tests, 0 failed, 6 errored, 1 disabled, 0 warning(s)
 
**Expected behavior**
Should be successfully run as in another schema in the same database.

**Example code**
betwnstr/test_betweenstr packages from demo project

**Additional context**
Maybe this is grant / object related (eg. duplicate system views in the schema or similar)

jgebal:
Fixed issue with utPLSQL failing to run when schema contains package named the same as schema owning it.
It is now possible to have utPLSQL test package like this:
```
create or replace package some_schema.some_schema as
  --%suite

  --%test
  procedure sample_test;
end;
/

create or replace package body some_schema.some_schema as
   procedure sample_test is begin ut.expect(1).to_equal(1); end;
end;
/
```

The tests from the package can be invoked by calling:
`exec ut.run('some_schema.some_schema');`

If only schema is provided, all test packages in schema are executed.
`exec ut.run('some_schema');`

Resolves #569
Resolves #885

### DIFF
--- a/source/core/ut_metadata.pkb
+++ b/source/core/ut_metadata.pkb
@@ -21,31 +21,6 @@ create or replace package body ut_metadata as
   g_cached_object varchar2(500);
   ------------------------------
   --public definitions
-
-  procedure do_resolve(a_owner in out nocopy varchar2, a_object in out nocopy varchar2, a_procedure_name in out nocopy varchar2) is
-    l_name          varchar2(200);
-    l_context       integer := 1; --plsql
-  begin
-    l_name := form_name(a_owner, a_object, a_procedure_name);
-    do_resolve(l_name,l_context,a_owner,a_object, a_procedure_name);
-  end do_resolve;
-
-  procedure do_resolve(a_fully_qualified_name in varchar2,a_context in integer,a_owner out nocopy varchar2, a_object out nocopy varchar2, 
-    a_procedure_name out nocopy varchar2) is
-    l_dblink        varchar2(200);
-    l_part1_type    number;
-    l_object_number number;
-  begin
-    dbms_utility.name_resolve(name          => a_fully_qualified_name
-                             ,context       => a_context
-                             ,schema        => a_owner
-                             ,part1         => a_object
-                             ,part2         => a_procedure_name
-                             ,dblink        => l_dblink
-                             ,part1_type    => l_part1_type
-                             ,object_number => l_object_number);
-  end;
-  
   function form_name(a_owner_name varchar2, a_object varchar2, a_subprogram varchar2 default null) return varchar2 is
     l_name varchar2(200);
   begin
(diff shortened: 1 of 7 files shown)


source/core/ut_metadata.pkb
```diff
@@ -327,12 +283,16 @@ create or replace package body ut_metadata as
...
-    ut_metadata.do_resolve(a_full_object_name,7,l_schema,l_object, l_procedure_name);
-    return l_object;
+    l_result := regexp_substr(
+      a_full_object_name,
+      '^([A-Za-z0-9$#_]+|".*?")\.([A-Za-z0-9$#_]+|".*?")', subexpression => 2
+      );
+    if not l_result like '"%"' then
+      l_result := upper(l_result);
+    end if;
+    return l_result;
```
lwasylow: I would do here dbms_assert to verify that we have a correct name returned in case regexp result got name not right.

jgebal: good catch

PR #940, opened 2019-06-16
Fixed issue with parsing via trigger source with leading comments

jgebal:
Resolves problem when sources are passed via JDBC or dynamic SQL for compilation and source code contains leading comments.
The code was adjusted to properly read sources even when they are getting passed with leading comments containing create package statement.
Now, the following will be properly parsed:
```sql
  begin
    execute immediate q'[
      /*
      * Some multiline comments before package spec
      create or replace package dummy_test_package dummy comment to prove that we pick the right piece of code
      */
      -- create or replace package dummy_test_package dummy comment to prove that we pick the right piece of code
      --Some single-line comment before package spec
      create or replace package dummy_test_package as
        --%suite(dummy_test_suite)
        --%rollback(manual)

        --create or replace package dummy_test_package dummy comment to prove that we pick the right piece of code
        
        --%test(dummy_test)
        --%beforetest(some_procedure)
        procedure some_dummy_test_procedure;
      end;]';
  end;
/
```

### DIFF
--- a/source/core/annotations/ut_annotation_manager.pkb
+++ b/source/core/annotations/ut_annotation_manager.pkb
@@ -267,10 +267,11 @@ create or replace package body ut_annotation_manager as
       l_sql_text    ora_name_list_t := a_sql_text;
     begin
       if a_parts > 0 then
-        l_sql_text(1) := regexp_replace(l_sql_text(1),'^\s*create(\s+or\s+replace){0,1}(\s+(editionable|noneditionable)){0,1}\s+{0,1}', modifier => 'i');
         for i in 1..a_parts loop
           ut_utils.append_to_clob(l_sql_clob, l_sql_text(i));
         end loop;
+        l_sql_clob := ut3.ut_utils.replace_multiline_comments(l_sql_clob);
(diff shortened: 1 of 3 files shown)


source/core/annotations/ut_annotation_manager.pkb
```diff
@@ -267,10 +267,11 @@ create or replace package body ut_annotation_manager as
       l_sql_text    ora_name_list_t := a_sql_text;
```
lwasylow: Is there a need for that variable anymore its passed as attribute anyway.

lwasylow:
In build there is some odd log out during installation :
Synonym UT_RUN_INFO created.
bash: del: command not found
Installing component UT_TRIGGER_ANNOTATION_PARSING

source/core/annotations/ut_annotation_manager.pkb
jgebal: Good spot - fixed

jgebal:
The message comes from script: `define_ut3_owner_param.sql` I'm not sure how else to remove temp files using sqlpls for both Win and Unix/Mac

PR #949, opened 2019-06-20
Removed need to grant `ut_annotation_manager` to users.

jgebal:


source/api/ut_runner.pkb
```diff
@@ -264,5 +264,39 @@ create or replace package body ut_runner is
...
+        l_end_pos := instr(a_path,'.',l_start_pos);
+        if l_end_pos = 0 then
+          l_end_pos := length(a_path)+1;
+          l_at_end  := true;
+          end if;
+        l_item := substr(a_path,l_start_pos,l_end_pos-l_start_pos);
+        if l_item is not null then
+          l_result  :=
+            l_result ||
+              dbms_crypto.hash(
```
lwasylow: Would not be better to remove function `get_hash` from `ut_compound_data_helper` and put it into `ut_utils` then refer to that function rather than directly to  `dbms_crypto.hash`  that way it would be kept in one place?

jgebal: Ouch. Didn't notice the comment.
Can take care of it later.

Issue #975, opened 2019-07-15
Empty Annotation Cache After Installing utPLSQL 


PhilippSalvisberg:
**Describe the bug**

When I install utPLSQL in the development environment via `development/install.sh` then the annotation cache is empty for existing tests. 

**Provide version info**

Information about utPLSQL and Database version, 
```sql
SQL> set serveroutput on
SQL> declare
  2    l_version varchar2(255);
  3    l_compatibility varchar2(255);
  4  begin
  5    dbms_utility.db_version( l_version, l_compatibility );
  6    dbms_output.put_line( l_version );
  7    dbms_output.put_line( l_compatibility );
  8  end;
  9  /
19.0.0.0.0
19.0.0


PL/SQL procedure successfully completed.

SQL> select substr(ut.version(),1,60) as ut_version from dual;

UT_VERSION                                                  
------------------------------------------------------------
v3.1.8.3148-develop                                         

SQL> select * from v$version;

BANNER
--------------------------------------------------------------------------------
BANNER_FULL
------------------------------------------------------------------------------------------------------------------------------------
BANNER_LEGACY                                                                        CON_ID
-------------------------------------------------------------------------------- ----------
Oracle Database 19c Enterprise Edition Release 19.0.0.0.0 - Production           
Oracle Database 19c Enterprise Edition Release 19.0.0.0.0 - Production                                                              
Version 19.2.0.0.0                                                                                                                  
Oracle Database 19c Enterprise Edition Release 19.0.0.0.0 - Production                    0


SQL> select * from nls_session_parameters;

PARAMETER                      VALUE                                                           
------------------------------ ----------------------------------------------------------------
NLS_LANGUAGE                   AMERICAN                                                        
NLS_TERRITORY                  SWITZERLAND                                                     
NLS_CURRENCY                   SFr.                                                            
NLS_ISO_CURRENCY               SWITZERLAND                                                     
NLS_NUMERIC_CHARACTERS         .'                                                              
NLS_CALENDAR                   GREGORIAN                                                       
NLS_DATE_FORMAT                DD.MM.RR                                                        
NLS_DATE_LANGUAGE              AMERICAN                                                        
NLS_SORT                       BINARY                                                          
NLS_TIME_FORMAT                HH24:MI:SSXFF                                                   
NLS_TIMESTAMP_FORMAT           DD.MM.RR HH24:MI:SSXFF                                          
[...]

PR #986, opened 2019-08-05
Cleanup and refactoring of annotation cache.

jgebal:
Added ability to refresh annotation cache with DDL trigger enabled if schema was not full-scanned.
Resolves: #975

With this change, re-install of utPLSQL framework will no longer cause issues on installations using DDL trigger for improved framework startup time.


jgebal:
@PhilippSalvisberg - that should be OK for the issue you've reported.

jgebal:
I did some more exploratory testing.
Seems like this version doesn't work well with DDL trigger disabled.
I'll need to spend some more time on rework.

PR #995, opened 2019-09-20
Corrected rebuild_annotation_cache parameter in the Install Guide

abasharin:
Parameter name should be `a_object_owner`.
See [package specification](https://github.com/utPLSQL/utPLSQL/blob/48b9d2e1d8cce7cce714bb6b0d1816788c4eb3a8/source/api/ut_runner.pks#L85).


jgebal:
Thank you @abasharin 
Fixes like this one really help make the project even better.
Every smallest contribution matters.

👍

PR #1008, opened 2019-10-20
Added ability to exclude items by tags.

### LINKED ISSUE #983: Allow to run utPLSQL without certain tags (ut.run(a_tags=>'-notThisTag'))
**Is your feature request related to a problem? Please describe.**
Playing around with tags, for me it is often much easier to tag several problematic (or long-running) tests which I *don't* want to run in several situtations.

**Describe the solution you'd like**
Would be great to enable ut-runner Tags-parameter to allow `-tag`:

```
begin
  ut.run(a_tags=>'-long_running');
end;
```
This would run all tests which don't have the `long_running` tag.

```
begin
  ut.run(a_tags=>'wip,-long_running');
end;
```
This would run all tests that have the `wip` tag but not the `long_running` tag

**Additional context**
We discussed an alternative approach with logic AND/OR operators (`a_tags => 'tag1&tag2|tag3`) but the `-` seems much more useful.

jgebal:
Resolves #983

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ env:
     #utPLSQL released version directory
     - UTPLSQL_DIR="utPLSQL_latest_release"
     - SELFTESTING_BRANCH=${TRAVIS_BRANCH}
-    - UTPLSQL_CLI_VERSION="3.1.7"
+    - UTPLSQL_CLI_VERSION="3.1.8"
     # Maven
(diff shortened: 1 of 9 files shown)


pesse:
Failing SonarCloud is due to the coverage bug in cli 3.1.7
Will provide 3.1.8 today, latest tomorrow (some things I still want to address)

PR #1059, opened 2020-03-16
Feature/swap develop schema

jgebal:
The change contains:
- Swapping schema for development so that `ut3` contains release version of framewok and `ut3_develop` contains the schema under developments
- Change how `ut3_info` context is named when deploying framework to different schema
- Fixes to cache handling for `ut_suite_cache`


PhilippSalvisberg:
What is the reason for this change?

Issue #77, opened 2016-10-17
Add be_within 'fuzzy' matcher


jgebal:
For numbers we should allow a comparison with a percent_off deviation.
The syntax:
`ut.expect( a_actual =>1.23 ).to_( be_within( a_pct => 1, a_expected => 1.23456789 ) )` 
should give success
`ut.expect( a_actual =>1.2 ).to_( be_within( a_pct => 1, a_expected => 1.23456789 ) )` 
should give success

Calculation logic.
a_actual between (a_expected - ((a_expected \* a_pct)/100) ) and (a_expected + ((a_expected \* a_pct)/100) )

jgebal:
Similar should be available for dates, timestamps and intevals (once suported)

Pazus:
Rescheduled to version 3.1

proldan:
What should `pct` mean in dates and intervals?

Shoelace:
that may not make a whole lot of sense..
more useful for dates would be if it meant "number of seconds difference"
ie so 1-jan-2017 10:12:13  would match 1-jan-2017 10:12:14

i guess you could say 50%  would mean +/- 12hours

jgebal:
I think that for dates, timestamps and intervals we should not think of within in terms of PCT.
For those data-types we should use interval as a measure of allowed deviation.
so the syntax woule be like:
```sql
ut.expect(sysdate).to_( be_within( interval '1' day, to_date( '20170515', 'yyyy-mm-dd' ) ) );
```

MickeVr:
I customized my utplsql implementation to be able to compare floats. However, instead of a relative precision, I used an absolute precision, but it sounds like the same idea as I already applied.  Would it be an idea to make this absolute / relative distinction in a parameter?

Shoelace:
that could work  so instead of
`ut.expect( a_actual =>1.23 ).to_( be_within( a_pct => 1, a_expected => 1.23456789 ) )`
you could also have
`ut.expect( a_actual =>1.23 ).to_( be_within( a_amt => 0.123, a_expected => 1.23456789 ) )`

that works for date/timestamp/interval comparison better then a_pct also  and keeps it somewhat consistent to a date/timpstamp comparitson could be 
`ut.expect( a_actual =>SYSDATE ).to_( be_within( a_amt => INTERVAL '1' WEEK, a_expected => DATE '2018-10-17' ) )`

Shoelace:
[...]

PR #1076, opened 2020-06-07
Feature/rework matcher

lwasylow:
Adding a matcher within_of. 
Rework some of the compound expectation to use a pass a matcher to expectation to remove methods from ut_equal_compound.
Implements #77


jgebal:
@lwasylow - sorry for not reviewing it yet. 
I'll look at it this week.

test/ut3_user/expectations/binary/test_to_be_within.pks
```diff
@@ -0,0 +1,31 @@
+create or replace package test_to_be_within is
```
jgebal, line 1: Tests for `not_to_be_within` are missing and so we can't say if the behavior works ok for negated matcher.

docs/userguide/expectations.md
```diff
@@ -333,6 +333,7 @@ The matrix below illustrates the data types supported by different matchers.
 |       **be_like**       |      |         |  X   |      |        |           |                               |                                        |    X     |                                 |                                 |        |                             |        |      |
 |      **be_empty**       |  X   |         |  X   |      |        |           |                               |                                        |          |                                 |                                 |   X    |              X              |        |  X   |
 |     **have_count**      |      |         |      |      |        |           |                               |                                        |          |                                 |                                 |   X    |              X              |        |  X   |
+| **be_within().of_()**   |      |         |      |   x  |   x    |           |                               |                                        |          |                                 |                                 |   X    |              X              |        |  X   |
```
jgebal, line 336: This is not correct.
The matcher doesn't support cursor/json/nested tables

source/expectations/matchers/ut_be_within.tpb
```diff
@@ -0,0 +1,113 @@
...
+  end;
+  
+  member procedure of_(self in ut_be_within, a_expected number) is
+    l_result ut_be_within := self;
+  begin 
+    l_result.expected := ut_data_value_number(a_expected);
+    l_result.expectation.to_(l_result );        
+  end;
+  
+  member procedure of_(self in ut_be_within, a_expected date) is
```
jgebal, line 51: Seems like the following datatypes are missing:
- timestamp
- timestamp with timezone
- timestamp with local timezone 

We could also add support for `intervals`, so users could have:
```sql
  ut.expect( interval '2' days ).to_be_within(interval '1' second').of_(interval '2' days);
```

source/expectations/matchers/ut_be_within.tpb
```diff
@@ -0,0 +1,113 @@
...
+      l_result := (self as ut_matcher).run_matcher(a_actual);
+    end if;
+    return l_result;
+  end;
+
+  overriding member function failure_message(a_actual ut_data_value) return varchar2 is
+    l_distance varchar2(32767);
+  begin
+    l_distance := case 
+                    when self.dist is of (ut_data_value_number) then
```
jgebal, line 93: I suspect that this CASE isn't needed.
I would expect the `to_string` method of the actual type to be invoked without need to use `treat`

source/expectations/matchers/ut_be_within.tpb
```diff
@@ -0,0 +1,113 @@
...
+                    when self.dist is of (ut_data_value_dsinterval) then
+                      treat(self.dist as ut_data_value_dsinterval).to_string 
+                    else
+                      null
+                    end;
+                    
+    return (self as ut_matcher).failure_message(a_actual) || ' '||l_distance ||' of '|| expected.to_string_report();
+  end;
+
+  overriding member function failure_message_when_negated(a_actual ut_data_value) return varchar2 is
```
jgebal, line 106: Shouldn't we calculate the distance here too? 
It's not covered by tests so hard to say

source/expectations/matchers/ut_be_within.tpb
```diff
@@ -0,0 +1,113 @@
...
+    l_result ut_be_within := self;
+  begin 
+    l_result.expected := ut_data_value_date(a_expected);
+    l_result.expectation.to_(l_result );    
+  end;
+  
+  overriding member function run_matcher(self in out nocopy ut_be_within, a_actual ut_data_value) return boolean is
+    l_result boolean;
+  begin
+    if self.expected.data_type = a_actual.data_type then
```
jgebal, line 61: This code is pretty hard to read. Lot of treat calls and nested if statements.

There must be a way to make this code more readable.

source/expectations/ut_expectation.tpb
```diff
@@ -719,5 +694,37 @@ create or replace type body ut_expectation as
     self.not_to( ut_contain(a_expected).negated() );
   end;
   
+  member function  to_be_within(a_dist natural) return ut_be_within is
+    l_result ut_matcher;
```
jgebal, line 698: use `l_result ut_be_within;`
that way there is no need for `return treat(... as )`

source/expectations/ut_expectation_base.tpb
```diff
@@ -0,0 +1,44 @@
+create or replace type body ut_expectation_base as
```
jgebal, line 1: I wonder if both `ut_expectation_base` as well as `ut_matcher_base` are required.
Normally, you cannot have a circular dependency and to overcome that one of type needs to have a `_base` and the other must depend on `_base`.

I would need to have a look in depth to see if the solution would works with single base type.
It would also be good to figure out pros/cons of having `ut_machter_base` vs `ut_expectation_base` to choose the one that fits better.

We might consider the `_base` type to be an interface. Something Oracle types don't have really.
It would definitely be an interface if it would be marked as non instantiable.

source/expectations/ut_expectation_base.tps
```diff
@@ -0,0 +1,25 @@
...
+  limitations under the License.
+  */
+  actual_data         ut_data_value,
+  description         varchar2(4000 char),
+
+  --base matcher executors
+  member procedure to_(self in ut_expectation_base, a_matcher ut_matcher_base),
+  member procedure not_to(self in ut_expectation_base, a_matcher ut_matcher_base)
+) not final
+/
```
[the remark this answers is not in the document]
jgebal: Make it `not final not instantiable`

source/expectations/ut_expectation_base.tps
```diff
@@ -0,0 +1,25 @@
...
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+  */
+  actual_data         ut_data_value,
+  description         varchar2(4000 char),
+
+  --base matcher executors
+  member procedure to_(self in ut_expectation_base, a_matcher ut_matcher_base),
```
jgebal, line 22: If we want to prevent those methods from being overridable they should be marked as `final`

source/expectations/ut_expectation_base.tpb
```diff
@@ -0,0 +1,44 @@
...
+    l_matcher       ut_matcher := treat(a_matcher as ut_matcher);
+    l_message       varchar2(32767);
+  begin
+    l_expectation_result := coalesce( l_matcher.run_matcher_negated( self.actual_data ), false );
+
+    l_message := coalesce( l_matcher.error_message( self.actual_data ), l_matcher.failure_message_when_negated( self.actual_data ) );
+    ut_expectation_processor.add_expectation_result( ut_expectation_result( ut_utils.to_test_result( l_expectation_result ), self.description, l_message ) );
+  end;
+end;
+/
```
jgebal, line 44: newline

source/expectations/matchers/ut_equal.tpb
```diff
@@ -172,6 +172,18 @@ create or replace type body ut_equal as
...
+  member procedure include(self in ut_equal, a_items varchar2) is
+  begin
+    include( ut_varchar2_list( a_items ) );
+  end;
+
+  member procedure include(self in ut_equal, a_items ut_varchar2_list) is
+    l_result ut_equal := self;
+  begin
+    l_result.options.include.add_items(a_items);
+    l_result.expectation.to_(l_result );
```
jgebal, line 184: I think we need some sort of exception handling.
The syntax that is will work perfectly well are:
- `ut.expect( l_refcursor ).to_( equal(l_refcursor).include( ... ) );`
- `ut.expect( l_refcursor ).to_equal(l_refcursor).include( ... );`

It is however possible to call:
`exec equal(l_refcursor).include( ...);`
In this case, the code will fail dramatically even though it is possible to make a call like this.

Instead of throwing an un-handled exception we should throw a user-friendly one.
Something like:
`ORA-20000 - utPLSQL matcher can only be called as part of expectation. Instead of calling equal( ... ).include( ... ); , call ut.expect( ... ).to_( equal( ... ).include( ... ) );`

Not sure what exact error should be but definitely something that can help users understand their mistake.

source/expectations/matchers/ut_equal.tpb
```diff
@@ -186,13 +198,32 @@ create or replace type body ut_equal as
...
+  member procedure exclude(self in ut_equal, a_items varchar2) is
+  begin
+    exclude( ut_varchar2_list( a_items ) );
+  end;
+
+  member procedure exclude(self in ut_equal, a_items ut_varchar2_list) is
+    l_result ut_equal := self;
+  begin
+    l_result.options.exclude.add_items(a_items);
+    l_result.expectation.to_(l_result );
```
jgebal, line 210: all calls to: `l_result.expecatation` in this type are assuming that the `expectation` attribute is set.

jgebal: Given the number of places this is called, it would be good to delegate exception handling to a helper procedure.

docs/userguide/expectations.md
lwasylow: Oops must edited incorrectly MD file..

docs/userguide/expectations.md
```diff
@@ -1093,6 +1094,87 @@ SUCCESS
...
+begin
+  ut.expect(9).to_be_within_pct(10).of_(10);
+end;
+/
+```
+
+**Example 3.**
+```sql
+begin
+  ut.expect(sysdate).to_be_within_pct(interval '1' day).of_(sysdate + 1);
```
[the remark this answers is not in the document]
lwasylow: Oops, it was late. The dates are not supported for pct. Good spot

docs/userguide/expectations.md
```diff
@@ -1093,6 +1094,87 @@ SUCCESS
...
+/
+```
+
+Returns following output via DBMS_OUTPUT:
+```
+Failures:
+ 
+  1) wihtin_test
+      Actual: 3 (number) was expected to be within 1 of 5 (number)
+      at "UT3_DEVELOP.UT_BE_WITHIN.OF_", line 48 l_result.expectation.to_(l_result );
```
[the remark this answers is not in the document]
lwasylow: No that's should not happen. Silly error I think

source/expectations/matchers/ut_be_within.tpb
lwasylow: I'm not sure if that's right. It's a hit against logic as interval words itself describe the distance and have no meaning without of relative point. Distonace of the distance feel counter intuitive to me.

jgebal: What if I use interval as a column to measure duration.
Ant I expect some duration to be 1 hour +/- 5 minutes?
This seems perfectly valid.
In that sense we're saying it sould take 1 hour with 5 minutes tolerance.

jgebal: I guess this discussion should be in issue rather than in PR

source/expectations/ut_expectation_base.tpb
lwasylow: Unfortunately the matcher have expectation as attribute and expectation is using matcher to call to_ method.

docs/userguide/expectations.md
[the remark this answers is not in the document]
lwasylow: Thats interesting its only for new matcher. Need to investigate.

pesse:
What's missing here? Maybe I can help

lwasylow:
> What's missing here? Maybe I can help

Thats a bit of a can of worm. 
I will go back to it had a hell of half year, up for my ears:)

jgebal:
Would be nice to have this PR refreshed.
Let me start by merging the branch from develop to kick off a fresh build on github actions

source/expectations/matchers/ut_be_within_helper.pkb
```diff
@@ -0,0 +1,46 @@
...
+    a_value_1 ut_data_value, a_value_2 ut_data_value, a_distance ut_data_value
+  ) return boolean is
+    l_result integer;
+  begin
+    execute immediate q'[
+            begin
+              :result :=
+                case
+                  when
+                    treat(:a_value_1 as ut3_develop.]'||a_value_1.self_type||q'[).data_value
```
jgebal, line 29: We cannot use hardcoded schema name in dynamic SQL like that.
Schema name will be different on different installations.
Is schema name really needed at all? The package is created with authid definer and so schema resolution should work OK (in schema where the package exists)

jgebal:
While running the feature tests locally just now, I have noticed a pitfall of using `interval` datatype as a time distance.
The below block fails with exceptions:
```sql
set serverout on;
begin
  ut.expect(date '2022-01-30').to_be_within(INTERVAL '1'  MONTH).of_(date '2022-01-30');
end;
/
begin
  ut.expect(date '2022-05-31').to_be_within(INTERVAL '1'  MONTH).of_(date '2022-05-31');
end;
/
```
```
begin
  ut.expect(date '2022-01-30').to_be_within(INTERVAL '1'  MONTH).of_(date '2022-01-30');
end;
Error report -
ORA-01839: date not valid for month specified
ORA-06512: at line 11
ORA-06512: at "UT3.UT_BE_WITHIN_HELPER", line 24
ORA-06512: at "UT3.UT_BE_WITHIN", line 114
ORA-06512: at "UT3.UT_EXPECTATION_BASE", line 26
ORA-06512: at "UT3.UT_BE_WITHIN", line 44
ORA-06512: at line 2
01839. 00000 -  "date not valid for month specified"
*Cause:    
*Action:

Error starting at line : 6 in command -
begin
  ut.expect(date '2022-05-31').to_be_within(INTERVAL '1'  MONTH).of_(date '2022-05-31');
end;
Error report -
ORA-01839: date not valid for month specified
ORA-06512: at line 3
ORA-06512: at "UT3.UT_BE_WITHIN_HELPER", line 24
ORA-06512: at "UT3.UT_BE_WITHIN", line 114
ORA-06512: at "UT3.UT_EXPECTATION_BASE", line 26
ORA-06512: at "UT3.UT_BE_WITHIN", line 44
ORA-06512: at line 2
01839. 00000 -  "date not valid for month specified"
*Cause:    
*Action:
```

This is because:
January 30 + Interval '1' month, according to Interval maths is **February 30th** 😞 
and 
May 31st + Inter

val '1' month, according to Interval maths is **June 31st** 

So, we would really need to rethink the maths and meaning of + 1 month to be more like `add_months` logic which assumes that if the day of month falls beyond last day of a month, it means it should be last day of a month.
Example:
```sql
select add_months(date '2022-05-31',1) from dual;
2022-06-30 12:00:00 AM
```

It seems that only the `interval month` logic suffers from that problem.

jgebal:
Well. `interval year` can also suffer from that problem on leap years.
https://stackoverflow.com/questions/35695175/oracle-db-ora-01839-date-not-valid-for-month-specified-29-02-2016-leap-year

mathewbutler:
The stack overflow link references this spec:

http://www.inf.fu-berlin.de/lehre/SS05/19517-V/FolienEtc/sql-foundation-aug94.pdf

Table 11— “Valid values for fields in datetime items” (page 130) states:

DAY - “ Within the range 1 to 31, but further constrained by the value of MONTH and YEAR fields, according to the rules for well-formed dates in the Gregorian calendar.”

This suggests that valid Gregorian calendar days should be the result. 

I’d this behaviour of INTERVAL types an Oracle bug?

lwasylow:
> Well. `interval year` can also suffer from that problem on leap years. https://stackoverflow.com/questions/35695175/oracle-db-ora-01839-date-not-valid-for-month-specified-29-02-2016-leap-year

I think we can work this around as we already have function to extract string from interval. I reckon we can try go this route to translate `interval` into `add_months` etc.

mathewbutler:
> The stack overflow link references this spec:
> 
> http://www.inf.fu-berlin.de/lehre/SS05/19517-V/FolienEtc/sql-foundation-aug94.pdf
> 
> Table 11— “Valid values for fields in datetime items” (page 130) states:
> 
> DAY - “ Within the range 1 to 31, but further constrained by the value of MONTH and YEAR fields, according to the rules for well-formed dates in the Gregorian calendar.”
> 
> This suggests that valid Gregorian calendar days should be the result.
> 
> I’d this behaviour of INTERVAL types an Oracle bug?

More digging and Oracle state this not a bug - just the way INTERVAL works. There’s an asktom from 2005 - Mr Kyte covers it all:

https://asktom.oracle.com/pls/apex/f?p=100:11:0::::p11_question_id:1157035034361

jgebal:
I think I've found a solution that would still rely on internal oracle logic related to intervals.
We currently use the below pseudo-code
```sql
declare
  l_result boolean;
begin
  l_result := 
    a_actual  between (a_expected - a_distance) and (a_expected + a_distance);
end;
```
The above results in problems when we calculate: `(a_expected -/+ a_distance)` for `a_distance` expressed as YM Interval when the calculation would result in invalid day of month like below:
```sql
select (timestamp '2022-01-30 14:45:56.123456'  + interval '0-1' year to month) from dual;
```

However we can express the same intent with a different formula:
```sql
declare
  l_result boolean;
begin
  l_result := ( least( a_actual, a_expected ) - greatest(a_actual, a_expected) ) year to month <= a_distance;
end;
```
That way, we avoid the problems with invalid dates rather than adding/subtracting interval to date, we calculate YM distance between dates and then compare it to expected max distance.

For `date` datatype of expected & actual this would look like below:
```sql
declare
  l_result boolean;
begin
  l_result := ( cast(least( a_actual, a_expected ) as timestamp) - cast( greatest(a_actual, a_expected) as timestamp) ) year to month <= a_distance;
end;
```

jgebal:
I have implemented required fixes.
@lwasylow @mathewbutler  - would you mind having a look at latest commit?

jgebal:
TODO - review readme and documentation update to assure that the matcher is well explained along with it's limitations and behaviour to make sure there is no confusion when folks will use it.

source/expectations/matchers/ut_equal.tpb
jgebal: @lwasylow 
I don't really remember why this change was done.
I think it's part of some clean-up operation but not really part of new matcher. Is your memory any better than mine?

lwasylow: @jgebal 

Went back to old conversations on slack from may 27th 2020 and found this:

_To not duplicate old methods from matchers we can change behaviour.
UT_MATCHER is additional attribute of expectation which is set when function is called.
Procedure TO_EQUAL in UT_EXPECTATION works the same._

jgebal: Yeah. I've found it. But after 2 years I can't even understand what I had in mind.
I'll put the original (polish) version here, just in case, as the free version of SLACK that we use purges messages history.
>Propozycja:
Żeby nie duplikować wszystkich metod z matcherów w ut_Expectation  można zmienić to zachowanie w następujący sposób:
ut_matcher a teraz dodatkowy atrybut expectation który jest ustawiany gdy wołana jest funcja.
Procedura to_equal w ut_Expectation  działa jak dotychczas - raportuje expecattaion
Funkcja to_equal  w ut_Expecattion_compound :
ustawia wartość pola expectation w przekazanym matcher  na wartość self
zwraca matcher, tak żeby metchod chaining teraz już używał metod matchera (dzięki temu nie trzeba już duplikować dodatkowych metod w ut_expectation
Matcher może odwołać się teraz do expectation z którego został wywołany.
W procedurze matchera (ostatnia metoda w łańcuchu) self.expecation is null  matcher nie woła expecataion, w innym wypadku procedury  matcher wołaję metodę to_ przekazując self 
W ten sposób składnia:
ut.expect(...).to_( equal(...).incude(...) ) sprawia, że:
matcher nie woła expectation, bo expectation nie jest ustawione w matcher
matcher jest przekazany do expectation i expectation wykonuje sprawdzenie wartości i raportuje
Składnia:
ut.expect(...).to_equal(...).include sprawia że:
to_equal ustawia self jako expectation w matcherze, zwraca matcher i dokonuje się metchod chaining
include  wywoływane jest na matcherze i poniewarz jest to procedura i pole expectation jest ustawione, matcher woła metodę  ecpectation.to_  przekazując self jako arcument
Powinno działać, ale jest 2 w nocy więc mogę majaczyć

source/expectations/matchers/ut_equal.tpb
jgebal: We can deal with this later.

jgebal:
**Question:**
Should the `to_be_within` matcher allow for comparison of 
- timestamp vs timestamp with time zone
- timestamp vs timestamp with local time zone
- date vs timestamp 
- etc.

That is, should we relax the strict datatype check that is present for the `to_equal` and allow for more relaxed check?
I think now is the time for this discussion before we merge and release this new matcher.

jgebal:
I have fixed 2 issues with pct matcher:
1. divisor=0 on expected = 0
2. matcher worked only one way.
`ut.expect(99).to_be_within_pct(1).of_(100)` was giving SUCCESS
`ut.expect(101).to_be_within_pct(1).of_(100)`  was giving FAILURE

With the change introduced now we will have:
```
begin
  ut3_develop.ut.expect(99).to_be_within_pct(1).of_(100);
  ut3_develop.ut.expect(101).to_be_within_pct(1).of_(100);
  ut3_develop.ut.expect(99).to_be_within_pct(1).of_(100);
  ut3_develop.ut.expect(101).to_be_within_pct(1).of_(100);
  ut3_develop.ut.expect(0).to_be_within_pct(1).of_(0);
  ut3_develop.ut.expect(0).to_be_within_pct(100).of_(1);
  ut3_develop.ut.expect( 1.0001 ).to_be_within_pct( -0.01 ).of_(1);
end;
/

SUCCESS
  Actual: 99 (number) was expected to be within 1 % of 100 (number)
SUCCESS
  Actual: 101 (number) was expected to be within 1 % of 100 (number)
SUCCESS
  Actual: 99 (number) was expected to be within 1 % of 100 (number)
SUCCESS
  Actual: 101 (number) was expected to be within 1 % of 100 (number)
SUCCESS
  Actual: 0 (number) was expected to be within 1 % of 0 (number)
SUCCESS
  Actual: 0 (number) was expected to be within 100 % of 1 (number)
SUCCESS
  Actual: 1.0001 (number) was expected to be within -.01 % of 1 (number)

```
@lwasylow

jgebal (approved):

PR #1171, opened 2021-12-06
Fix typo in advanced_data_comparison.md

kukimik:
Typo (is -> if). This way the sentence makes some sense to me. Hope I got it right.

### DIFF
--- a/docs/userguide/advanced_data_comparison.md
+++ b/docs/userguide/advanced_data_comparison.md
@@ -250,8 +250,9 @@ FAILURE
   at "anonymous block", line 15
 ```
 **Note**
-> `join_by` matcher is much faster on performing data comparison, consider using `join_by` over unordered
-> `contain` matcher is not considering order of compared data-sets. Using `unordered` makes no difference (it's default)
+> Consider using `join_by( columns... )` over `unordered()` with the `equal` matcher. The `join_by` method is much faster at performing data comparison.
+>
+> The `contain` matcher is not considering the order of the compared data-sets. Using `unordered` makes no difference (it's default).
 
 
 ## Join By option


kukimik:
Then I'd say the sentence is not clear enough; I'm not able to parse it, and even after your comment I'm not sure what is the meaning of this note. Perhaps it can be reworded?

lwasylow:
I guess it could be reworded to say that  when using a contain matcher, the order of data is irrelevant and they are compared similar way when unordered matcher is used.
Certainly the word `if` is not a right one to use it in this scenario in my opinion.

kukimik:
I think I got it. The two lines:

> `join_by` matcher is much faster on performing data comparison, consider using `join_by` over unordered

and 

> `contain` matcher is not considering order of compared data-sets. Using `unordered` makes no difference (it's default)

are two separate sentences. The first line is missing a period at the end. Am I right?

If so, please check the updated commit.

docs/userguide/advanced_data_comparison.md
```diff
@@ -250,8 +250,9 @@ FAILURE
   at "anonymous block", line 15
 ```
 **Note**
-> `join_by` matcher is much faster on performing data comparison, consider using `join_by` over unordered
-> `contain` matcher is not considering order of compared data-sets. Using `unordered` makes no difference (it's default)
+> Consider using `join_by` over unordered `equal`, as the `join_by` matcher is much faster at performing data comparison.
```
jgebal, line 253: Consider using `join_by( columns... )` over `unordered()` with `to_equal` matcher. The `join_by` method is much faster at performing data comparison.

jgebal (changes requested):

docs/userguide/advanced_data_comparison.md
kukimik: Corrected. I just changed "with `to_equal` matcher" to "with the `equal` matcher", because in other parts of the docs (e.g. in [`expectations.md`](https://github.com/utPLSQL/utPLSQL/blob/develop/docs/userguide/expectations.md#equal)) the matchers are named without the `to_` prefix.

However, now I wonder whether the part "with the `equal` matcher" is necessary at all.

jgebal (approved):

PR #1173, opened 2021-12-22
Minor typo, style and grammar fixes

kukimik:
This should be less problematic than #1171, I mostly fix obvious typos.

I believe the new note on negating `to_contain` is in place, as it may be rather surprising that `not_to_contain` is not a negation of `to_contain`.

### DIFF
--- a/docs/userguide/expectations.md
+++ b/docs/userguide/expectations.md
@@ -24,7 +24,7 @@ Expectation is a combination of:
 - the expected value
 - optional custom message for the expectation 
 - the matcher used to perform comparison
-- them matcher parameters (actual value), depending on the matcher type
+- the matcher parameters (actual value), depending on the matcher type
   
 
 Matcher defines the comparison operation to be performed on expected (and actual) value.
@@ -88,7 +88,7 @@ There are two ways to use expectations:
 - without invoking the utPLSQL framework - running expectations standalone
 
 ## Running expectations within utPLSQL framework
-When expectations are ran a part of test suite, the framework tracks:
+When expectations are ran as a part of a test suite, the framework tracks:
 - status of each expectation 
 - outcomes (messages) produced by each expectation
 - call stack to each expectation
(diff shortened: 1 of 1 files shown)


docs/userguide/expectations.md
```diff
@@ -312,6 +312,9 @@ FAILURE
 ```
 Since NULL is neither *true* nor *false*, both expectations will report failure.
 
+**Note:**
+Beware that `not_to_contain` is not a simple negation of `to_contain`: `not_to_contain` checks whether the actual and the expected set are disjoint.
```
lwasylow, line 316: I don't think that note is necessary. Matcher was written to be as close in plain English as possible and I think `not_to_contain` describe its function.

docs/userguide/expectations.md
```diff
@@ -916,13 +919,13 @@ To change the behavior of `NULL = NULL` comparison pass the `a_nulls_are_equal =
 
 This matcher supports only compound data-types comparison. It check if the actual set contains all values of expected subset.
 
-When comparing data using `contain` matcher, the data-types of columns for compared compound types must be exactly the same.
+When comparing data using the `contain` matcher, the data-types of columns for compared compound types must be exactly the same.
 
-The matcher supports all advanced comparison options as `equal` like: `include` , `exclude`, `join_by` etc..
+The matcher supports all advanced comparison options supported by `equal` like: `include` , `exclude`, `join_by` etc..
```
lwasylow, line 924: It does not sounds good as word support is repeated twice in same sentence. Not sure if change is required here.

lwasylow (commented):
Most

docs/userguide/expectations.md
kukimik: I am not sure if the following construct is correct English: *'to support all Xs as Y'*. Maybe it's just my ignorance, but perhaps *The matcher supports the same advanced comparison options as `equal`* would be better?

For now, I reverted it to the original form.

docs/userguide/expectations.md
kukimik: I believe `not_to_intersect` describes its function much better. A set `A` does not contain a set `B` if and only if there exists an element of `B` that is not an element of `A`; in particular the sets do not have to be disjoint. I know what I am saying, I am a mathematician. :wink:

The first line of the section says: _Expectations provide a very convenient way to perform a check on a negated matcher._ However, `not_to_contain` is not a negation of `to_contain`.  

Anyway, I removed the note. If you change your mind, it can always be included later.

lwasylow: I think `not to intersect` would be a more accurate name indeed but maybe less plain. We tried to stick to plain English as possible so idea is that set A not contain set B.

lwasylow (approved):
Changes looks good to me.

jgebal:
Thanks for all the improvements

Issue #1098, opened 2020-09-04
fails when a grand child nested type have NULL


hirerajan:
**Describe the bug**
Lets say you have type of an object with three layers, i.e grand parent, parent and child type. 
for example m_package_tab

m_package_obj( a varchar2(2), b number, t_parent_tab( p_ame varchar2(100), c_child_tab(name varchar2,age number)))

and lets say for our data, our child_tab we have null data,
and when you compare the data using following, you will get an error saying **Fetched column value is NULL**

ut.expect(anydata.convertcollection(l_package_tab)).to_equal(anydata.convertcollection(l_test_package_tab)).unordered;

OUTPUT:
ORA-01405: fetched column value is NULL
      ORA-06512: at "UTP3.UT_DATA_VALUE_REFCURSOR", line 336
      ORA-06512: at "SYS.DBMS_CRYPTO_FFI", line 159
      ORA-06512: at "SYS.DBMS_CRYPTO", line 86
      ORA-06512: at "UTP3.UT_UTILS", line 868
      ORA-06512: at line 1
      ORA-06512: at "UTP3.UT_DATA_VALUE_REFCURSOR", line 331
      ORA-06512: at "UTP3.UT_DATA_VALUE_REFCURSOR", line 384
      ORA-06512: at "UTP3.UT_DATA_VALUE_ANYDATA", line 129
      ORA-06512: at "UTP3.UT_EQUAL", line 229
      ORA-06512: at "UTP3.UT_EXPECTATION", line 26
      ORA-06512: at "UTP3.UT_EXPECTATION_COMPOUND", line 146

**Workout Around**
       ut.expect( xmltype( anydata.convertCollection(l_package_tab  ) ).getClobVal() ).to_equal( xmltype(anydata.convertCollection(l_test_package_tab) ).getClobVal() );

There is also problem with workout Around as the order of the output does not match. I cannot use **.unordered** with perhaps three layer nested loop or with xmltype. Not sure why. 

utPLSQL version : v3.1.10.3349

pesse:
Hey @hirerajan , could you provide a complete example?
This seems to be a rather complex setup, you could assist us greatly with looking into it

hirerajan:
@pesse ,
Execuse my UT, for the shake of this testing, I quickly wrote following program that I was able to replicate.  I am directly using table data instead of setting up test data on unit test. 

Let me know if this helps. 

[...]

Issue #1083, opened 2020-06-21
Unreadable & misleading results when comparing nested objects


jgebal:
**Describe the bug**
When comparing nested objects, the report produced is confusing and unreadable.

**Provide version info**
Information about utPLSQL and Database version, 

DB version `18.0.0`
`utPLSQL v3.1.11.3392-develop`

**Information about client software**
not relevant

**To Reproduce**
Execute the below script
```sql
create or replace type obj as object(
  value varchar2(200)
);
create or replace type obj_lst as table of obj;

create or replace type test_obj as object (
  id number,
  lst obj_lst
);

declare
  l_expected test_obj;
  l_actual   test_obj;
begin
  l_expected := test_obj(1, obj_lst(obj('1'),obj('2'),obj('3'),obj('2')));
  l_actual   := test_obj(1, obj_lst(obj('1'),obj('3'),obj('3'),obj('2')));
  ut.expect(anydata.convertObject(l_actual)).to_equal(anydata.convertObject(l_expected));
end;
/
```

The output is:
```
FAILURE
  Actual: ut3_develop.test_obj was expected to equal: ut3_develop.test_obj
  Diff:
  Rows: [ 1 differences ]
    Row No. 1 - Actual:   <LST><OBJ><VALUE>1</VALUE></OBJ><OBJ><VALUE>3</VALUE></OBJ><OBJ><VALUE>3</VALUE></OBJ><OBJ><VALUE>2</VALUE></OBJ></LST><OBJ><VALUE>3</VALUE></OBJ><OBJ><VALUE>2</VALUE></OBJ><OBJ><VALUE>3</VALUE></OBJ><OBJ><VALUE>1</VALUE></OBJ><OBJ><VALUE>3</VALUE></OBJ><OBJ><VALUE>3</VALUE></OBJ><OBJ><VALUE>1</VALUE></OBJ><OBJ><VALUE>2</VALUE></OBJ><OBJ><VALUE>1</VALUE></OBJ><OBJ><VALUE>3</VALUE></OBJ><OBJ><VALUE>3</VALUE></OBJ>
[...]

Issue #1082, opened 2020-06-21
Failure when comparing nested objects


jgebal:
**Describe the bug**
When comparing nested objects the comparison fails with an exception.

**Provide version info**
`utPLSQL 3.1.10` - any database

**Information about client software**
not applicable

**To Reproduce**

Run the below script
```sql
create or replace type some_value as object(
  the_value   number,
  is_relative integer(1)
);

create or replace type some_item as object (
  first_value  some_value,
  second_value some_value
);

declare
  l_expected some_item;
  l_actual   some_item;
begin
  l_expected := some_item(some_value( 1, 0 ), some_value(2.22, 1));
  l_actual   := some_item(some_value( 1, 0 ), some_value(2.22, 0));
  ut.expect(anydata.convertObject(l_actual)).to_equal(anydata.convertObject(l_expected));
end;
/
```
See the exception:
```
ORA-00902: invalid datatype
ORA-06512: at "UT3.UT_DATA_VALUE_REFCURSOR", line 336
ORA-06512: at "UT3.UT_COMPOUND_DATA_HELPER", line 628
ORA-06512: at "UT3.UT_DATA_VALUE_REFCURSOR", line 328
ORA-06512: at "UT3.UT_DATA_VALUE_REFCURSOR", line 384
ORA-06512: at "UT3.UT_DATA_VALUE_ANYDATA", line 129
ORA-06512: at "UT3.UT_EQUAL", line 229
ORA-06512: at "UT3.UT_EXPECTATION", line 26
ORA-06512: at "UT3.UT_EXPECTATION", line 96
ORA-06512: at line 7
```

**Expected behavior**

The expectation should report data differences.

**Additional context**

The behavior seem to show up when the object contains nested objects.

PR #1179, opened 2022-01-26
Feature/1082 failure when comparing nested objects

lwasylow:
Fixing issue where code was failing on the nested objects due to "object" type being pulled into dynamic SQL as datatype to extract into. 

Fixing issue when null value was passed to get_hash function causing compare to fail 

Address a possible scenario when more than 2 levels of nesting for same objects caused ambiguity of the column name.

Closing #1082 
Closing #1098 
Closing #1083

### DIFF
--- a/test/ut3_user/expectations/test_expectation_anydata.pks
+++ b/test/ut3_user/expectations/test_expectation_anydata.pks
@@ -229,7 +229,7 @@ create or replace package test_expectation_anydata is
   procedure nested_varray_diff_data; 
   
   --%test ( Comparision won't fail on user_defined type that is null as per issue 1098 )
-  procedure user_defined_type_null_issue_1098;
+  procedure user_def_type_null_issue_1098;
   
   --%test ( Reports success when comparing complex nested objects )
   procedure complex_nested_object_success;      
--- a/test/ut3_user/expectations/test_expectation_anydata.pks
+++ b/test/ut3_user/expectations/test_expectation_anydata.pks
@@ -206,6 +206,9 @@ create or replace package test_expectation_anydata is
   
   --%test ( Failure of comparing nesting objects )
   procedure failure_nesting_objects;
+ 
+  --%test ( Failure of comparing double nested objects ) 
+  procedure failure_double_nested_objects;
   
 end;
 /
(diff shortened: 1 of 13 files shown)


simasch (approved):

jgebal:
Let me have a deeper look into that PR before I approve it

test/ut3_user/expectations/test_expectation_anydata.pks
```diff
@@ -203,5 +203,29 @@ create or replace package test_expectation_anydata is
  
   --%test ( Empty Array not equal array with space )
   procedure arr_empty_nqua_arr_e_unord;  
+  
+  --%test ( Failure of comparing nested objects )
```
jgebal, line 207: Can we name those tests better?
Maybe something like:
```
--%test(Correctly reports a diff on failure with nested object)
--%test(Correctly reports a diff on failure with double nested object)
--%test(Reports success when comparing identical nested object)
--%test(Reports success when comparing identical double nested object)
--%test(Correctly reports a diff on failure with nested object list)
```

test/ut3_user/expectations/test_expectation_anydata.pks
```diff
@@ -203,5 +203,29 @@ create or replace package test_expectation_anydata is
...
+  procedure failure_nested_object_list;
+ 
+  --%test ( Success of comparing nested object list )
+  procedure success_nested_object_list;
+  
+  --%test(Nested VARRAYS with same data)
+  procedure nested_varray_same_data;
+  
+  --%test(Nested VARRAYS with different data)
+  procedure nested_varray_diff_data;
```
jgebal, line 229: Can you add a test to cover for regression of #1098 (on nested element being null)?

jgebal (changes requested):

jgebal:
I've reviewed the PR. It addresses all 3 issues.
Brilliant work @lwasylow :ok_hand: 👍

test/ut3_user/expectations/test_expectation_anydata.pks
lwasylow: Good point.

test/ut3_user/expectations/test_expectation_anydata.pks
lwasylow: Thats test replicated using our types and regression created.

jgebal (approved):

PR #1183, opened 2022-02-06
Feature/add reason to disable

### LINKED ISSUE #610: Add support for `reason` in the `--%disabled` annotation
It would be of great benefit if we would support the disabled annotations with reason.
So that instead of specifying `--%disabled` users could specify `--%disabled(reason)`
The reason should show on reports along the disabled indication.
It is to be considered if the reason should be inline with test/suite description or if it should be placed on the next line under the test/suite with proper indentation and coloring.

lwasylow:
Adding option to put a reason for disabling tests.
CLOSES #610


source/core/ut_suite_builder.pkb
```diff
@@ -412,6 +413,12 @@ create or replace package body ut_suite_builder is
     end if;
     l_test.disabled_flag := ut_utils.boolean_to_int( l_proc_annotations.exists( gc_disabled));
 
+    if l_proc_annotations.exists(gc_disabled) then
+      l_annotation_texts := l_proc_annotations( gc_disabled);
+      --take the last definition if more than one was provided    
+      l_test.disabled_reason := l_annotation_texts(l_annotation_texts.first);
```
jgebal, line 419: In line 573 you trim the text while here it is not trimmed.

source/core/ut_suite_builder.pkb
```diff
@@ -560,8 +567,13 @@ create or replace package body ut_suite_builder is
     if a_annotations.by_name.exists(gc_tags) then
       add_tags_to_suite_item(a_suite, a_annotations.by_name(gc_tags),a_suite.tags);
     end if;
+    
     a_suite.disabled_flag := ut_utils.boolean_to_int(a_annotations.by_name.exists(gc_disabled));
-
+    if a_annotations.by_name.exists(gc_disabled) then
+      l_annotation_text := trim(a_annotations.by_name(gc_disabled)(a_annotations.by_name(gc_disabled).first));
```
jgebal, line 573: Is trimming needed? Value is trimmed here, but not trimmed in line 419.
Can we make this consistent? 
No need to use `l_annotation_text` variable. Value can be assigned directly to `a_sute.disabled_reason`

jgebal (commented):
We need to also update other reporters except `ut_documentation_reporter`
Particularly: 
- `ut_junit_reporter` - tag `<skipper>reason</skipped>`
- `ut_tfs_junit_reporter` - tag `<skipper>reason</skipped>`

For the `ut_realtime_reporter` - It would be good to check with @PhilippSalvisberg 
For the `ut_teamcity_reporter` - I'm not sure if the format supports a disabled/skipped description. The format description can be found [here](https://confluence.jetbrains.com/display/TCD9/Build+Script+Interaction+with+TeamCity). 
I didn't run utPLSQL on TeamCity for a long time and don't have local setup ready. 
You would need to install and run TeamCity build locally (docker?) to actually test it.

PhilippSalvisberg:
> We need to also update other reporters except `ut_documentation_reporter` Particularly:
> (...)
> For the `ut_realtime_reporter` - It would be good to check with @PhilippSalvisberg 

I don't think this information is required for the realtime reporter. If you decide to add it, it will not have any effect. The SQLDev extension and the PL/SQL Developer plugin will ignore this additional information. If you think the behaviour in the IDE should change then I suggest to open dedicated issues for SQLDev and PL/SQL Developer.

jgebal:
@lwasylow @simasch 
Not sure what are your thoughts on having disabled reason visible in SQL/PLSQL Developer IDE.
I think it would be a good to have a complete feature and similar behaviour across all use-cases & IDEs.

lwasylow:
> @lwasylow @simasch Not sure what are your thoughts on having disabled reason visible in SQL/PLSQL Developer IDE. I think it would be a good to have a complete feature and similar behaviour across all use-cases & IDEs.

So personally I think if we want to have that feature enabled it makes sense to be consistent.  I think it's useful piece of information. 
I don't think it should hold up release of that feature as we can stage it given a dependency between objects and time needed for @PhilippSalvisberg  @simasch to give us specs ( and I mean their personal time  :) )  but would ultimate aim for that to be included?

PhilippSalvisberg:
> > @lwasylow @simasch Not sure what are your thoughts on having disabled reason visible in SQL/PLSQL Developer IDE. I think it would be a good to have a complete feature and similar behaviour across all use-cases & IDEs.
> 
> So personally I think if we want to have that feature enabled it makes sense to be consistent. I think it's useful piece of information. However lets hold on with making that call to see if other reporters support the reason of disabling. I don't think it should hold up release of that feature as we can stage it given a dependency between objects and time needed for @PhilippSalvisberg @simasch to give us specs ( and I mean their personal time :) ) but would ultimate aim for that to be included?

As mentioned before, you can include it if you want. The SQLDev extension and the PL/SQL Dev plugin will work without any change since the just pick the information form the XML events they need. If you want the IDEs I suggest to open dedicated issues. I think the commercial tools TOAD and DataGrip do not use the realtime reporter. However, if the reason for a disabled test should be displayed in the test runner they need to change something as well.

lwasylow:
Please see snippet from junit reporter.
image

lwasylow:
Here are results for `ut_tfs_junit_reporter`
image

as per [documentation](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/test/publish-test-results?view=azure-devops&tabs=junit%2Cyaml)

<html>
<body>


Error message | /Testsuites/testsuite/testcase/failure.Attributes["message"].Value Or /Testsuites/testsuite/testcase/error.Attributes["message"].Value Or /Testsuites/testsuite/testcase/skipped.Attributes["message"].Value
-- | --



</body>
</html>

simasch:
If created a feature request for the PL/SQL Developer plugin https://github.com/utPLSQL/utPLSQL-PLSQL-Developer/issues/25

lwasylow:
Here results for `ut_teamcity_reporter`
image

jgebal (approved):

PR #1186, opened 2022-02-09
Feature - include/exclude objects and schemas by regex

### LINKED ISSUE #1053: Exclude Objects from code coverage by wildcard
**Is your feature request related to a problem? Please describe.**
We have a database schema which includes temporary objects - this means objects which are required only for a short time like procedure for upgrades or objects which are still work in progress - and legacy test code. Temporary objects are always prefixed with the username of the author like:
windowsusr_my_table
Test Procedures are prefixed with test_ and cannot be moved to utPLSQL because they do an integration test rather than a unit test as they access and read data from remote systems and therefore need manual checks of the data.

**Describe the solution you'd like**
We would like to exclude objects using a wildcard like:
windowsusr_*
test_*

**Describe alternatives you've considered**
I tried to list every object which currently exists and which should be excluded but gave up after about 15 names.

**Additional context**
We use Oracle 12.1 with Sqldeveloper and installed the utPLSQL plugin.

lwasylow:
Adding support for inclusion and exclusion of objects and schemas using regex.

```sql
begin
  ut.run(
    a_path => 'unit_test_schema', 
    a_reporter => ut_coverage_html_reporter(), 
    a_include_schema_expr => '(user_...)|(..._user)',
    a_include_object_expr => '(bonus.*)|(.*_process_.*)',
    a_exclude_schema_expr => '(user_t..)|(t.._user)',
    a_exclude_object_expr => '(.*award.*)|(.*calc.*)'
  );
end;
```

Current approach is that `a_include_object_expr ` and `a_include_schema_expr ` silently overwrites `a_include_object_expr `

Closes #1053


source/core/coverage/ut_coverage.pkb
```diff
@@ -94,6 +96,15 @@ create or replace package body ut_coverage is
...
              and s.type  = f.object_type
              and s.owner = f.object_owner';
+    elsif a_coverage_options.include_schema_expr is not null or a_coverage_options.include_object_expr is not null then
+      l_full_name := q'[lower(s.type||' '||s.owner||'.'||s.name)]';
+      if a_coverage_options.include_schema_expr is not null then
+        l_filters := q'[and regexp_like(s.owner,']'||a_coverage_options.include_schema_expr||q'[','i')]';
+      end if;
+      
+      if a_coverage_options.include_object_expr is not null then
+        l_filters := l_filters|| q'[ and regexp_like(s.name,']'||a_coverage_options.include_object_expr||q'[','i')]';
```
jgebal, line 106: This expression cannot be concatenated as it opens up a security hole for SQL injection.
We need to use a bind variable to pass that value to the filtered query.

source/core/coverage/ut_coverage.pkb
```diff
@@ -30,11 +30,12 @@ create or replace package body ut_coverage is
   end;
 
   function get_cov_sources_sql(a_coverage_options ut_coverage_options, a_skip_objects ut_object_names) return varchar2 is
-    l_result                varchar2(32767);
+    l_result                varchar2(32767); --QUESTION: Given fact that we can pass large regex filters same as objects filter should we consider a clob ?
```
jgebal, line 33: Let's keep the variables as they are.
Let's pass the filters as parameters instead of concatenating them into the SQL query

source/core/coverage/ut_coverage.pkb
```diff
@@ -111,13 +122,25 @@ create or replace package body ut_coverage is
         end;
     end if;
 
+    
+    if a_coverage_options.exclude_schema_expr is not null then
+       l_regex_exc_filters := l_regex_exc_filters||q'[ and not regexp_like(s.owner,']'||a_coverage_options.exclude_schema_expr||q'[,'i')]';
```
jgebal, line 127: This needs to be passed as a parameter to the SQL query.

source/core/coverage/ut_coverage.pkb
```diff
@@ -84,8 +85,9 @@ create or replace package body ut_coverage is
                where s.owner = el.owner and  s.name = el.name
            )
        and line > 0
+       {regex_filters}
```
jgebal, line 88: Why was this added as a separate filter at the end instead of being added to the main filter in top WITH clause?

lwasylow: Good point. I have to review some things after your recent PR to improve COVECOV sql anyway.

source/core/types/ut_coverage_options.tps
```diff
@@ -21,13 +21,21 @@ create or replace type ut_coverage_options force as object (
   exclude_objects  ut_object_names,
   include_objects  ut_object_names,
   file_mappings    ut_file_mappings,
+  include_schema_expr varchar2(32767),
```
jgebal, line 24: Change to varchar2(4000) to keep the type SQL-compatible.
Otherwise, we cannot safely use this type in SQL

Issue #1192, opened 2022-02-25
Extend ut_realtime_reporter to include disabled reason


PhilippSalvisberg:
## Is your feature request related to a problem? Please describe

In utPLSQL v3.1.12 a feature was added to add a reason for a disabled test. See PR #1183 and issue #610. Some reporters were extended, but this information is missing in `ut_realtime_reporter`. The utPLSQL runner for SQL Developer and PL/SQL Developer are using the `ut_realtime_reporter`. So this reporter should include the reason for a disabled test. This is a prerequisite to display the information in the UI.

## Describe the solution you'd like

Add the node `disabledReason` in the `test` node that is produced in [`before_calling_run`](https://github.com/utPLSQL/utPLSQL/blob/v3.1.12/source/reporters/ut_realtime_reporter.tpb#L45). Use CamelCase as in other nodes of this reporter. Examples: `executableType`, `ownerName`, `objectName` and `procedureName`. 

## Describe alternatives you've considered

Snake-case naming of the node `disable_reason` as suggested by @jgebal in utPLSQL/utPLSQL-PLSQL-Developer#25 . The naming would be inconsistent with the other nodes in the `ut_realtime_reporter`. Therefore I strongly suggest to use CamelCase for this additional node. I suggest to use `disabledReason`.

## Additional context

The UI related issues are utPLSQL/utPLSQL-PLSQL-Developer#25 and utPLSQL/utPLSQL-SQLDeveloper#143 .

PR #1193, opened 2022-02-25
Added disabledReason to `ut_realtime_reporter`

jgebal:
The `ut_realtime_reporter` used by utPLSQL-SQL-Developer-extension and utPLSQL-PLSQL-Developer-plugin  is now capable of providing reason for a disabled test.
The reason is provided as a tag `<disabledReason>reason</disabledReason>` in the `pre-run` event, describing the suites before run is started.

If no reason is provided, the tag is not present in the output.

Resolves: #1192


lwasylow (approved):

PhilippSalvisberg (approved):
Looks good. Thanks a lot @jgebal 👍

PR #1203, opened 2022-03-31
Add ability to run tests by part of a name.

### LINKED ISSUE #470: Add ability to run tests by part of a name.
Currently you need to specify a full name to execute test procedure / suite / package
Example:
```sql
exec ut.run('owner.test_package');
--or
exec ut.run('owner.test_package.test_procedure');
```
We could add ability to use '*' at the end of the package name to run all tests/suites by part of the name.
So I could call:
```sql
exec ut.run('owner.test_p*');
--or
exec ut.run('owner.test_package.test_pr*');
```

Adding the '*' at the end of the name should be most easy and most beneficial.

Allowing '*' in any place of name would be more natural, but there is very little value in it and implementation of such feature would be a bigger effort.

### LINKED ISSUE #1148: Is there a way I can run only specific tests (test packages starting with XX%, etc)?
When we run ut_runner.run, is there a way we can get it run test packages starting with XX% in a particular schema? Basically, we have a schema called APPS which contains thousands of objects and it takes lot of time to identify the annotations on all these objects.

lwasylow:
Introducing a capability to call tests by part of name.
We can call out using path or object name / procedure name.
Schema name has to be defined.
This will allow to simplify a syntax for multiple tests / suites e.g.
```
begin
 ut3_develop.ut.run( 'ut3:*_package*');
end;
/
```
or
```
begin
 ut3_develop.ut.run( 'ut3.test_package_*');
end;
/
```

Closes #470 
Closes #1148


jgebal (commented):
I see a problem with this PR (besides the specific comments)

Only 329 tests get executed when this branch is used as source for installing into UT3 schema. So it looks like only tests from one schema are executed. [See here](https://github.com/utPLSQL/utPLSQL/runs/5786138663?check_suite_focus=true#step:23:519)

That indicates there is a regression bug/problem that is not captured by existing unit tests.

lwasylow:
> I see a problem with this PR (besides the specific comments)
> 
> Only 329 tests get executed when this branch is used as source for installing into UT3 schema. So it looks like only tests from one schema are executed. [See here](https://github.com/utPLSQL/utPLSQL/runs/5786138663?check_suite_focus=true#step:23:519)
> 
> That indicates there is a regression bug/problem that is not captured by existing unit tests.

Thats interesting, good spot... 
I will check it out, there are two places where I this can go wrong here, either expand paths or get cached suits sql.

lwasylow:
> I see a problem with this PR (besides the specific comments)
> 
> Only 329 tests get executed when this branch is used as source for installing into UT3 schema. So it looks like only tests from one schema are executed. [See here](https://github.com/utPLSQL/utPLSQL/runs/5786138663?check_suite_focus=true#step:23:519)
> 
> That indicates there is a regression bug/problem that is not captured by existing unit tests.

So that was caused by sorting of paths before passing to `reconstruc_cache` . There were scenarios were a level 2 could appear right after level 3 but been then followed by level 3 again. This should not be an issue with a single sequential check however in current situation I have changed a code to order by siblings. I also slightly modified place where ordering is happening however I believe the sorting by siblings get a better results than nls_sort.

source/core/ut_suite_cache_manager.pkb
```diff
@@ -102,158 +87,262 @@ create or replace package body ut_suite_cache_manager is
...
+         where c.tags multiset intersect l_exclude_tags is not empty
+       )
+       select value(c) as obj
+       bulk collect into  l_suite_tags
+         from table(a_suite_items) c
         where exists (
           select 1 from included_tags t
-           where t.path||'.' like c.obj.path || '.%' /*all parents and self*/
-              or c.obj.path||'.' like t.path || '.%' /*all children and self*/
+           where t.path||'.' like c.path || '.%' /*all parents and self*/
```
jgebal, line 261:
```suggestion
           where t.path||'.' like c.path || '.%' /*all ancestors and self*/
```

source/core/ut_suite_cache_manager.pkb
```diff
@@ -102,158 +87,262 @@ create or replace package body ut_suite_cache_manager is
...
+              or c.path||'.' like t.path || '.%' /*all children and self*/
```
jgebal, line 262:
```suggestion
              or c.path||'.' like t.path || '.%' /*all descendants and self*/
```

source/core/ut_suite_cache_manager.pkb
```diff
@@ -102,158 +87,262 @@ create or replace package body ut_suite_cache_manager is
...
           )
         and not exists (
           select 1 from excluded_tags t
-           where c.obj.path||'.' like t.path || '.%' /*all children and self*/
-          )
-       ),]'
-           else
-             q'[dummy as (select 'x' from dual where :a_include_tag_list is null and :a_include_tag_list is null and :a_exclude_tag_list is null),]'
-           end;
+           where c.path||'.' like t.path || '.%' /*all children and self*/
```
jgebal, line 266:
```suggestion
           where c.path||'.' like t.path || '.%' /*all descendants and self*/
```

jgebal:
There are tests missing for:
- wildcard in beginning
- wildcard in the middle
- multiple wildcards

Documentation also does not mention that wildcard can be used in any part of path/object_name/procedure_name

So does it mean that wildcard is allowed only at the end of the search string or should the doc & tests be updated?

jgebal (approved):

Issue #1220, opened 2022-07-28
Uninstall script: buffer overflow error


kukimik:
**Describe the bug**

While using the `uninstall.sql` script I got the following error:

```
ORA-20000: ORU-10027: buffer overflow, limit of 2000 bytes
```

This happened in the "Dropping synonyms pointing to non-existing objects in schema UT3" part of the script (see `source/unistall_synonyms.sql`). The exception handler does not handle this gracefully, as it tries to write to the buffer which is already overflown:

```
    begin
      execute immediate 'drop '||syn.syn_name;
      dbms_output.put_line('Dropped '||syn.syn_name||' for object '||syn.for_object);
      exception
      when others then
      dbms_output.put_line('FAILED to drop '||syn.syn_name||' for object '||syn.for_object);
    end;
```

So the block exits and not all of the synonyms are dropped.

**Provide version info**
Oracle 12.1.0.2.0
utPLSQL v3.1.11

**Information about client software**
A rather old version of SQL*Plus

**Expected behavior**
All the synonyms are dropped and all the messages are logged.

jgebal:
Thank you for reporting

PR #1221, opened 2022-07-28
Fix uninstall scripts

jgebal:
Fixing problem with uninstall script buffer overflow on DBMS_OUTPUT.
Resolves: #1220

### DIFF
--- a/source/uninstall_objects.sql
+++ b/source/uninstall_objects.sql
@@ -1,5 +1,4 @@
 set echo off
-set serverout on
 declare
   procedure drop_if_exists(a_object_type varchar2, a_object_name varchar2) is
     l_count integer;


kukimik:
Note that with serveroutput turned off, failure to drop a synonym is likely to get unnoticed. See the error handling in `uninstall_synonyms.sql`: https://github.com/utPLSQL/utPLSQL/blob/fc5c4af28ad9416de894670a92cdc3fe9081f4f4/source/uninstall_synonyms.sql#L22, https://github.com/utPLSQL/utPLSQL/blob/fc5c4af28ad9416de894670a92cdc3fe9081f4f4/source/uninstall_synonyms.sql#L55

----------

By the way, also [this](https://github.com/utPLSQL/utPLSQL/blob/fc5c4af28ad9416de894670a92cdc3fe9081f4f4/source/uninstall_synonyms.sql#L18) `i := i + 1;` should be put after `execute immediate` succeeds. Maybe it can also be fixed here. Mentioning any failures in the summaries (` dbms_output.put_line(i||' synonyms dropped');`)  could also be helpful.

jgebal:
> Note that with serveroutput turned off,...

It won't as the output is enabled as unlimited inthis script.
https://github.com/utPLSQL/utPLSQL/blob/develop/source/uninstall.sql#L17
https://github.com/utPLSQL/utPLSQL/blob/develop/source/define_ut3_owner_param.sql#L17

PR #1224, opened 2022-08-08
Fix dropped synonyms counter.

kukimik:
The counter should not be incremented when an exception occurs.


jgebal (approved):

jgebal:
Thanks for your contribution @kukimik

PR #1225, opened 2022-08-15
Add a section on custom reporters.

kukimik:
Custom reporters are mentioned in a few places in the docs. Also, `README.md` lists "plugin architecture for reporters and matchers" as one of the key features. But there is no section in the docs that actually describes how to add a custom reporter. I believe that this short description of the process may be useful (it would be useful for me).

Sorry - I didn't have time to build the website  and check if it looks ok, but I don't see why it wouldn't.   (I've had a revolution in my family life in the last few months, and now I hardly have time for anything)

I decided not to describe the necessary grants when the custom reporter is created outside of the `UT3` schema. I state that this is not recommended, If someone wants to follow this path, let them do it at their own risk. Also, I'm not able to check it in different DB versions.

### DIFF
--- a/docs/userguide/reporters.md
+++ b/docs/userguide/reporters.md
@@ -161,5 +161,20 @@ Some of the information in debug log might be redundant.
 > - db object names
 > - etc.
 
- 
-       
+# Custom reporters
+
+It is possible to add your own reporters by creating an appropriate object type. In principle, it has to be a subtype of `ut_reporter_base`. However, if the reporter is expected to produce output consumable by a client oustside of the database (e.g. the data has to be reported to the screen or to a file), then you should base it on `ut_output_reporter_base` (which is a subtype of `ut_reporter_base`). In contrast, if you would like to create a reporter that, for example, saves the data to a database table, then it should be based directly on `ut_reporter_base`. (Currently, all reporters in the utPLSQL framework are based on `ut_output_reporter_base`.) Coverage reporters are based on `ut_coverage_reporter_base` (a subtype of `ut_output_reporter_base`).
+
+If need to produce a colored text output from the custom reporter, then you can build it basing on `ut_console_reporter_base` (a subtype of `ut_output_reporter_base`). In many cases it may also be more convenient to create the custom reporter type under a more specialized type, like `ut_documentation_reporter` or `ut_junit_reporter`, and override just some of the functionality.
+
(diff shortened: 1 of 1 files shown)


docs/userguide/reporters.md
```diff
@@ -161,5 +161,18 @@ Some of the information in debug log might be redundant.
 > - db object names
 > - etc.
 
- 
-       
\ No newline at end of file
+# Custom reporters
+
+It is possible to add your own reporters by creating an appropriate object type. In principle, it has to be a subtype of `ut_reporter_base`. However, in most cases, it is more convenient to create the custom reporter type under a more specialized type, like `ut_documentation_reporter` or `ut_junit_reporter`, and override just some of the functionality.
```
jgebal, line 166: We need to distinguish between `ut_reporter_base` which is a foundation for any reporter and 
`ut_output_reporter_base` which should be used as a base for reporter that is expected to produce output consumable by client (screen/file) outside of database.

Currently, all reporters in utPLSQL framework provide outputs to customer

If however you would like to create a reporter that saves data to a DB table for example, you should use `ut_reporter_base` 

utPLSQL will behave differently based on what type they are build on.

For reporters built on `ut_output_reporter_base` utPLSQL will consume output of the reporter and put it to screen before finishing the `ut.run()` call. For reporters not based on that type, output will not be consumed.

jgebal: You could build a reporter based on `ut_console_reporter_base` if you would like to produce a text output from reporter in different colors

jgebal:
@kukimik Thank you for contributions

jgebal:
Thank you for the contribution @kukimik 

I have added a suggestion to make the doc even better.
Please have a look at it whenever you have time.

docs/userguide/reporters.md
kukimik: Good points, thanks for the review. I'll improve the paragraph as suggested. I hope to do it over the weekend, but can't promise anything. I'm also going to mention where the core reporter types can be found in the repo.

kukimik:
@jgebal I try to contribute to the open source projects that I use. For me it's both a way to get better understanding of the tool, and to say "thank you" to the authors/maintainers.

docs/userguide/reporters.md
jgebal: Thanks Michał
I rally know how difficult it can be to find time & space to spend on writing some code outside of your da-to-day work. We appreciate all, even smallest contributions. They really make a difference.

kukimik: @jgebal I force-pushed an updated version.

PR #1231, opened 2022-08-31
Add explicit grants for public packages

barsema:
The privileges on the referenced packages are often revoked from PUBLIC as database hardening. Not relying on these grants being available makes the install script more robust.

See also issue #1229


jgebal:
Thank you @barsema 
This is a good addition to the jnstallation script

Issue #1232, opened 2022-09-05
ut_coverage_html_reporter crashes when code under test contain lines larger than 1911 characters


PhilippSalvisberg:
**Describe the bug**
When the source code contain one or more lines that are longer than 1911 characters (bytes?) then the `ut_coverage_html_reporter` crashes. 

**Provide version info**
```
21.0.0.0.0
21.0.0


PL/SQL procedure successfully completed.


UT_VERSION                                                  
------------------------------------------------------------
v3.1.13.4021-develop


BANNER                                                                           BANNER_FULL                                                                                                                                                      BANNER_LEGACY                                                                        CON_ID
-------------------------------------------------------------------------------- ---------------------------------------------------------------------------------------------------------------------------------------------------------------- -------------------------------------------------------------------------------- ----------
Oracle Database 21c Enterprise Edition Release 21.0.0.0.0 - Production           Oracle Database 21c Enterprise Edition Release 21.0.0.0.0 - Production                                                                                           Oracle Database 21c Enterprise Edition Release 21.0.0.0.0 - Production                    0
                                                                                 Version 21.3.0.0.0                                                                                                                                                                                                                                          



PARAMETER                      VALUE                                                           
------------------------------ ----------------------------------------------------------------
NLS_LANGUAGE                   AMERICAN                                                        
NLS_TERRITORY                  AMERICA                                                         
NLS_CURRENCY                   $                                                               
NLS_ISO_CURRENCY               AMERICA                                                         
NLS_NUMERIC_CHARACTERS         .,                                                              
NLS_CALENDAR                   GREGORIAN                                                       
NLS_DATE_FORMAT                DD.MM.YYYY HH24:MI:SS                                           
NLS_DATE_LANGUAGE              AMERICAN                                                        
NLS_SORT                       BINARY                                                          
NLS_TIME_FORMAT                HH.MI.SSXFF AM                                                  
NLS_TIMESTAMP_FORMAT           DD.MM.YYYY HH24:MI:SSXFF                                        

PARAMETER                      VALUE                                                           
------------------------------ ----------------------------------------------------------------
NLS_TIME_TZ_FORMAT             HH.MI.SSXFF AM TZR                                              
NLS_TIMESTAMP_TZ_FORMAT        DD.MM.YYYY HH24:MI:SSXFF TZR                                    
NLS_DUAL_CURRENCY              $                                                               
NLS_COMP                       BINARY                                                          
[...]

PR #1240, opened 2022-10-18
Ability to report coverage on long lines.

jgebal:
Sourcecode lines that result in more than 4000 chars after conversion to HTML should be reported successfully.
Resolves: #1232


PhilippSalvisberg (approved):
@jgebal looks good. Thank you very much!

PR #1261, opened 2023-07-07
fix(doc): Update running-unit-tests.md - remove example

gassenmj:
Just stumbled across a duplicate in Your (excellent!!) documentation. I think it can be removed. Thanks for utplsql <3

### DIFF
--- a/docs/userguide/running-unit-tests.md
+++ b/docs/userguide/running-unit-tests.md
@@ -118,7 +118,16 @@ begin
   ut.run(ut_varchar2_list('hr.test_apply_bonus','cust'));
 end;
 ```
-Executes all tests from package _hr.test_apply_bonus_ and all tests from schema _cust_.
+Executes all tests from package _hr.test_apply_bonus_ and all tests from schema _cust_ (passing individual items to be executed as elements of the ut_varchar2_list table type).
+
+
+```sql linenums="1"
+set serveroutput on
+begin
+  ut.run(ut_varchar2_list('hr.test_apply_bonus,cust'));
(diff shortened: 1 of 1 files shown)


jgebal:
Hi @gassenmj 
It is great to see that someone is actually reading the documentation 😄 

There is a slight difference n the two examples. and the documentation does not do a great job describing the differnece.


You can run tests by passing individual items to be executed as elements of the `ut_varchar2_list` table type.
```
set serveroutput on
begin
  ut.run(ut_varchar2_list('hr.test_apply_bonus','cust'));
end;

You can run tests by passing all items as a comma-separated-list of values into a single element of the `ut_varchar2_list` table type.
```
set serveroutput on
begin
  ut.run(ut_varchar2_list('hr.test_apply_bonus,cust)');
end;
```


See the difference:
```
begin
  ut.run(ut_varchar2_list('hr.test_apply_bonus','cust'));
  ut.run(ut_varchar2_list('hr.test_apply_bonus,cust)');
end;
```

The framework is very flexible in terms of format for this input parameter.

Perhaps better description of those examples would bring more value to the documentation.

What are your thoughts?

File: docs/userguide/running-unit-tests.md [hunk 1]
```diff
@@ -120,14 +120,6 @@ end;
 ```
 Executes all tests from package _hr.test_apply_bonus_ and all tests from schema _cust_.
 
-```sql linenums="1"
```
jgebal, line 123: Instead of deleting this example, a better description of difference between this and example above is needed.

PR #1276, opened 2024-05-21
Matcher shortcut syntax example fixed

bozo030:


pesse (approved):
Thanks for finding and fixing :)

jgebal (approved):

PR #1295, opened 2025-04-03
Enable duplicate detection on cursor comparison using join by

### LINKED ISSUE #1293: Cursor comparison not detecting duplicate rows
**Describe the bug**
When I compare two cursors using `.join_by` the duplicate rows are not detected (I'm attaching an example code).

**Provide version info**
Oracle Database 23ai Free Release 23.0.0.0.0
utPLSQL version: v3.1.14.4197
NLS: Polish
OS: x86_64/Linux 2.4.xx

Client: SQLDeveloper

**To Reproduce**
1. Prepare two identical cursors with a column you can join on.
2. Ensure one of the cursors has an additional row, identical to the ones already present.
3. Run `ut.expect` to compare them using `.join_by(...)`.
4. See a message (via `dbms_output`)  indicating a SUCCESS, yet clearly stating different rowcounts.

**Expected behavior**
The expectation should result in a FAILURE, not a SUCCESS. The rowcounts are clearly different, so the cursors don't yield the same rows, hence are not equal.

**Example code**
```sql
set serveroutput on
declare
  l_actual   sys_refcursor;
  l_expected sys_refcursor;
begin
  open l_expected for 
    select 'FOO'  username, 12 from dual union all
    select 'TEST' username, -600 user_id from dual
    order by 1 desc;
  open l_actual for 
    select 'FOO'  username, 12 from dual union all
    select 'TEST' username, -600 user_id from dual union all
    -- DUPLICATE!!!
    select 'TEST' username, -600 user_id from dual
    order by 1 asc;
  ut.expect( l_actual ).to_equal( l_expected ).join_by('USERNAME');
end;
/
--------
/* The result is:
SUCCESS
  Actual: refcursor [ count = 3 ] was expected to equal: refcursor [ count = 2 ]
*/
```

lwasylow:
Resolves #1293 
In certain scenarios when the data sets are identical but have an duplicate rows, the cursors join_by when doing join in SQL joined only by PK which resulted on full match from differences query.
Extended a PK join by to the duplicate number to make sure we match on the rows number.

Testing case was added however due to issue with a cursor comparison inside the logic the case will fail until merged into develop.

### DIFF
--- a/source/expectations/data_values/ut_compound_data_helper.pkb
+++ b/source/expectations/data_values/ut_compound_data_helper.pkb
@@ -300,8 +300,8 @@ create or replace package body ut_compound_data_helper is
       ut_utils.append_to_clob(a_partition_stmt,' row_number() over (partition by '||l_partition_tmp||' order by '||l_partition_tmp||' ) ');
       
       if a_pk_table.count > 0 then   
-        -- If key defined do the join or these and where on diffrences
-        a_join_by_stmt := ut_utils.table_to_clob(l_join_by_list, ' and ');
+        -- If key defined do the join or these and where on diffrences as well as on duplicate number when rows are same.
+        a_join_by_stmt := ' e."UT3$_Dup#No" = a."UT3$_Dup#No" and '||ut_utils.table_to_clob(l_join_by_list, ' and ');
       elsif a_unordered then
         -- If no key defined do the join on all columns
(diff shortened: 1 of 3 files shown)


jgebal:
Build is failing on 11 and 12 because procedure name is too long.
Until we officially drop support for 11, 12, we need main the compatibility.

lwasylow:
People getting use to good things to quickly :) Updated

jgebal (approved):

PR #1297, opened 2025-04-10
Creating a list of annotations allowed to treat package as annotated

### LINKED ISSUE #1278: Coverage not being reported when comment starts with --%param
I installed utPlSql on my computer. I am currently using Allround PL/SQL Developer. When I add a parameter definition as a comment like this -- %param into my source code (not a test suite package), utplsql checks this package and thinks as a test suite/package and inserts this info into the ut_suite_cache_package table.
When I run the test suite there is no problem. But when I want to see the coverage result, I cannot see the package which contains --%param keywords on the coverage result page. After inspecting ut_suite_manager.save_object_cache procedure, I see the **"a_suite_items IS NOT NULL AND a_suite_items.count = 0"** if clause but when I check the parameters, the a_suite_items parameter is null when the source code doesn't contain any %suite or others special keywords.
After updating the if clause like this **"IF a_suite_items IS NULL OR (a_suite_items IS NOT NULL AND a_suite_items.count = 0) THEN"** and compiles my non-test suite package and then runs the test suite, it deletes the related line from the ut_suite_cache_package table and shows the package name correctly on the coverage result page. 

@jgebal asked me to open this issue after a short conversation with him on Slack.

version info
v3.1.14.4197
Oracle Database 19c Enterprise Edition Release 19.0.0.0.0 - Production

lwasylow:
Resolve #1278 

Root problem is the issue that package is being treated as annotate which allows it to be pushed into cache_suite table.
The solution is the do not treat object as annotated if the `--%` does not contain a valid annotation that is specified by framework.
e.g. ignore comments like --%parameter

This will allow for greater flexibility for documenting.


jgebal:
This is not the correct way to solve this problem.
Annotation parsing must remain independent of suites.
This is by design.

There was an actual issue (A bug) in the suite cache management (not annotation cache)

I opened a separate PR to solve this problem. See #1313

PR #1305, opened 2025-07-31
Add TAP reporter

WayneNani:
I added a reporter for the [Test Anything Protocol ](https://testanything.org). It is easy to read for humans and machines alike and supported by a wide variety of consumers. These consumers can then aggregate test results from different projects or programming languages.

I tried to include all features from the [latest TAP Specification](https://testanything.org/tap-version-14-specification.html) as far as utPLSQL supports them (i.e., TODO tests are not supported in utPLSQL as far as I know). 

I also added some test cases that should provide a decent coverage.

Please feel free to let me know if I can add anything else.


lwasylow:
Hi @WayneNani,
Thank you very much for creating this.
I'm on holiday currently so will look when I'm back, possibly @jgebal will be able earlier.

One thing could you please update the documentation reporters section with screens etc.?

Kind regards Lukasz

WayneNani:
Hey @lwasylow ,
thanks for the quick reply! I changed the documentation as well, thanks for the heads-up.

WayneNani:
Hey @lwasylow and @jgebal ,

is there anything else I can do to streamline the process for you?

source/reporters/ut_tap_reporter.tpb
```diff
@@ -0,0 +1,124 @@
...
+    self.lvl := 0;
+    return;
+  end ut_tap_reporter;
+
+  member procedure print_comment(self in out nocopy ut_tap_reporter, a_comment clob) as
+  begin
+    self.print_clob(regexp_replace(a_comment, '^', '# ', 1, 0, 'm'));
+  end print_comment;
+
+  member function escape_special_chars(self in out nocopy ut_tap_reporter, a_string_to_escape clob) return clob as
```
lwasylow: Do you expect the strings here to be larger than 32k that we using CLOBs? I can see that you are using `testnames` etc there, so doesn't sound like needs `clob`.
If so for a sake of performance would suggest to change that to procedure with clob nocopy.
Running lot of temp clobs to for each line might not be best way,

jgebal:
Hello @WayneNani 

As you have probably noticed, the project was on pause for quite some time.
I am planning to give it a small refresh and so will aim to integrate your pull-request.

Will you be available to address the review comments (if any)?

WayneNani:
Hey @jgebal,

I am happy to address any remarks you have.

test/ut3_user/reporters/test_tap_reporter.pkb
```diff
@@ -0,0 +1,196 @@
+create or replace package body test_tap_reporter as
+
+  gc_boilerplate_suitepath_expression constant varchar2(300) := 'TAP version 14\s*1..1\s*# Subtest: org\s{5}1..1\s{5}# Subtest: utplsql\s{9}1..1\s{9}# Subtest: tests\s{13}1..1\s{13}# Subtest: helpers\s{17}1..1\s{17}# Subtest: A suite for testing different outcomes from reporters';
+
+
+  procedure compile_tests as
+    pragma autonomous_transaction;
+  begin
+
+    execute immediate q'[
```
jgebal, line 10: Please use a test suite with a suite path and with a context so that the tests document behaviour of nesting (TAP subtests).

jgebal:
@WayneNani 
One of widely used elements of utPLSQL is nested suites (with suitepaths) and contexts within suite.

How would the TAP reporter show results of nested suites or suite with contexts?

I would expect the report to show it as a subtest according to the TAP specification.

Even it it is already implemented in this PR, it is not mentioned in documentation of reporter and it is not expressed in the tests.

I would expect reporter to support the [Commented Subtests](https://testanything.org/tap-version-14-specification.html) (those are recommended).

image

source/reporters/ut_tap_reporter.tpb
```diff
@@ -0,0 +1,124 @@
...
+    self.print_text('1..' || a_run.items.count);
+    self.print_text(' ');
+  end before_calling_run;
+
+  overriding member procedure after_calling_run(self in out nocopy ut_tap_reporter, a_run in ut_run) as
+  begin
+    self.lvl := 0;
+  end;
+end;
+/
```
jgebal: In general, it's probably better to leave an empty newline at the end of each file.

test/ut3_user/reporters/test_tap_reporter.pks
```diff
@@ -0,0 +1,44 @@
...
+  procedure special_characters_in_deisabled_reason;
+
+  --%test(Don't escape special characters in comment)
+  procedure special_characters_in_comment;
+
+
+  --%afterall
+  procedure drop_help_tests;
+
+end test_tap_reporter;
```
jgebal, line 52: Include test(s) to document the behaviour with nested suites (Subtests).

jgebal (changes requested):

jgebal:
@WayneNani 
The PR is clean and implementation is really nice. 
Thank you for the contribution.
The only concerns is around support for nested suites and contexts.
It is be required to have that tested and documented.

Once this is done. We will include this change in next release (coming soon).

WayneNani:
@jgebal Thanks for the review! You are right, I forgot to include tests regarding suites and contexts and have fixed that. I decided to use the existing reporters test package which comes with a suitepath and context and left the escaping test package nearly untouched. 
I also included a test to still include the subtest even when it only consists of disabled tests.

jgebal:
Because our pipeline doesn't seem to work with external PRs anymore, I will pull and deploy your changes locally to take a look.

jgebal:
@WayneNani 
The PR looks really good.
Beautiful contribution tot he framework.

Thank you 👍 

Starting with next release we will be only suppoorting DB 19c and above.

Hopefully you don't need the reporter for any unsupported version of Oracle.

PR #1308, opened 2026-03-06
Add a_fail_on_error parameter to ut.run to support raising exceptions on test failures

MonikaMitura:
Description:
This PR adds the a_fail_on_error parameter to the run procedures in ut package.

Behavior:
If a_fail_on_error is set to true, the suite will raise an exception after the execution completes if one or more tests have failed or errored. This provides consistency with the existing functionality in ut_runner.run(a_fail_on_error => true).

Changes:
Added a_fail_on_error parameter to ut.run procedures.
Logic ensures that an exception is raised post-run if the test suite status is not successful.


lwasylow:
Hi @MonikaMitura  , can you branch out of utplsql repo?
Can you remove the .bak files ?
Can you create an appropriate tests for that ? 
Cheers

MonikaMitura:
1. Actually, I created my branch directly from the utPLSQL/utPLSQL develop
branch. Is there something specific in the history that looks off? Do you
want me to recreate the branch from scratch to ensure a clean commit
history?
2. Ok, I'll take care of that.
3. Could you please give me a hint on what kind of tests you're expecting
for this feature? Also, where is the best place to put them in the current
project structure?

lwasylow:
So it looks like you forked from utplsql repo rather than branch from our develop branch, I think the automatic tests don't run on forks.
As for tests please check this file : https://github.com/utPLSQL/utPLSQL/blob/develop/test/ut3_user/api/test_ut_run.pkb

Cheers

MonikaMitura:
Ok, I'll do it that way, probably on Monday.
I understand you want me to update the package and add new tests in
test_ut_run.pkb. I can do that. Thanks for the hint!

jgebal:
> So it looks like you forked from utplsql repo rather than branch from our develop branch, I think the automatic tests don't run on forks. As for tests please check this file : https://github.com/utPLSQL/utPLSQL/blob/develop/test/ut3_user/api/test_ut_run.pkb
> 
> Cheers

Hi Lukasz,
Only users that are designated as utPSLQL developers can create branches directly in our repo. The repo is protected from direct public development. This is by design.

Fork + PR from forked repo is the supported approach for external contributions.
As for tests and other stuff, I will look into this now.

@lwasylow  - thanks for quick feedback.

lwasylow:
Oh reall? Apologies :) 
I got so used to work on that repo that don't think used fork here before :)

source/api/ut.pkb
```diff
@@ -590,13 +605,21 @@ create or replace package body ut is
...
+        a_fail_on_errors         
       );
     end if;
     if l_reporter is of (ut_output_reporter_base) then
         treat(l_reporter as ut_output_reporter_base).lines_to_dbms_output();
     end if;
     raise_if_packages_invalidated();
+
+  exception when others then 
+    if l_reporter is of (ut_output_reporter_base) then
```
jgebal, line 618: Is there any specific reason to add this exception handler?
It seems to be out of scope of the change.

source/api/ut.pkb
```diff
@@ -253,7 +254,8 @@ create or replace package body ut is
     a_include_schema_expr varchar2 := null,
     a_include_object_expr varchar2 := null,
     a_exclude_schema_expr varchar2 := null,
-    a_exclude_object_expr varchar2 := null
+    a_exclude_object_expr varchar2 := null,
+    a_fail_on_errors boolean := false
```
jgebal: unit tests are required to confirm that the functionality is working.
Please add at least 1 tests to the `test_ut_run` package.
To confirm that when a failing UT is executed - an exception is thrown.

You could also add a negative test, to confirm that with null/false parameter value - no exception is thrown.

jgebal (commented):
Two small comments in code.

Note:
If you want to set up local DB for proper utPLSQL developmewnt - read the  [CONTRIBUTING,.md](https://github.com/utPLSQL/utPLSQL/blob/develop/CONTRIBUTING.md#prepare-environment-script)

source/api/ut.pkb
MonikaMitura: Test results won't show up in the output without this exception. 
without exception:
Zrzut ekranu 2026-03-09 101007

with exception:
Zrzut ekranu 2026-03-09 101230

jgebal:
@MonikaMitura 

I have talked with @PhilippSalvisberg some days ago about the root cause for this PR.

It seems that it is far more important to make a new `utPLSQL-cli` release so it can be used in enterprise and is not considered dead or legacy project.

So right now I will focus on creating a new release for utPLSQL-java-api and utPLSQL-cli to make them compatible with latest  java and give them a bit of refresh.


Some additional comments.
The unit tests that were added are quite an overkill. The changes in tests are not reflecting the changes in code.
The newly added test do not simply focus on testing the functionality of the change.
utPLSQL project file naming convention is to keep filenames in lowercase, not upper case.

PR #1310, opened 2026-03-08
Fix to the way the 23 handles the object duration in type plus image fixes.

lwasylow:
Updating builds and removing old versions of Oracle


lwasylow:
@jgebal Can you review and approve?
If you don't agree with a image changes we can take at least the reporter fix for the 23 error :)

source/core/types/ut_reporter_base.tpb
```diff
@@ -165,52 +165,53 @@ create or replace type body ut_reporter_base is
   end;
 
   overriding member procedure on_event( self in out nocopy ut_reporter_base, a_event_name varchar2, a_event_item ut_event_item) is
+    l_event_item ut_event_item := a_event_item;  -- ADD THIS LINE
```
jgebal: why is this change needed?

lwasylow: In the 23 apparently there is an issue with a duration of the object when casting/treat objects.  I think it's oracle bug and is patched in some versions that object is no longer in memory when not using in out.
It's probably something that oracle should fix but easier to do it here then rely on their patch

jgebal: I have worked around that problem in our pipeline by using version 23.26.0 instead of 23.26.1.

It is strange to see this exception especially as `a_event_item` is actually `in` not `in out`.

I am not a big fan of making changes in code to workaround Oracle bugs but if it solves the problem permanently than it's OK.

Since it is a solution for a problem, please create a separate issue for that specific error we see, so the error is documented in repository and it should be referenced in code comment.
That way there is clarity why we use this redundant and unnecessary assignment.

With this we minimize the risk of introducing a regression at some point in future, when we are no longer using Oracle version `23.26.1`  for builds and testing. Some users may still use it and be prone to the issue.

.github/workflows/build.yml
```diff
@@ -230,6 +247,12 @@ jobs:
           git config --global user.email "github-actions[bot]@users.noreply.github.com"
           git config --global user.name "github-actions[bot]"
 
+      - name: Setup Python for docs
```
jgebal, line 250: Is this needed? It seems like docs are generated just fine

source/core/types/ut_reporter_base.tpb
jgebal: I read your comment on #1310. This is resolving this problem then.
I will add a comment in code just to make it super clear.

jgebal:
```suggestion
    /* Resolves issue with ORA-21779 being thrown see issue: 1309 -> ( https://github.com/utPLSQL/utPLSQL/issues/1309#issuecomment-4020289898 ) */
    l_event_item ut_event_item := a_event_item;
```

jgebal (approved):

Issue #1309, opened 2026-03-07
Decommision support for 11, 12 and 18 databases.


lwasylow:
I think is time to decommision 11g builds and official support.

jgebal:
We should only keep 19 and above.
!9c goes out of premier support in 1 month. Extended support ends in 1 year.

https://www.itechtics.com/eol/oracle-database/

jgebal:
<img width="1190" height="892" alt="Image" src="https://github.com/user-attachments/assets/50ca4e75-437d-44b8-946d-4fbf96278845" />

jgebal:
I'm refreshing the github repo and b ringing the project back to stability now. I see that you also started to do the same.

lwasylow:
Yes @jgebal , almost there apart from 19c.
I think looking on the images there was never 19c XE. I wonder if there is an issue with that image

jgebal:
We always had our own DB builds images long before Docker was a thing that everyone had.
So we have our own images created.
For some reason, the 19c image that we have is not working well.
Maybe I need to rebuild it.
There is some strange issue with connecting to that image.
I can reproduce it locally.

Connections work well when I connect from my local Windows.
When I use Ubuntu 24 on Windows (WSL) , I get the same error as we see in pipeline.
When doing the same thing with Oracle 23 image - all works fine.

jgebal:
Well. I reported [the problem with Oracle 23 image](https://github.com/gvenzl/oci-oracle-free/issues/129) to Gerald - related to some non-deterministic error.
The other problem with our own 19c Image should now be resolved with [this commit](https://github.com/utPLSQL/utPLSQL/commit/71f4911ed3920bff2b6b01b9be6caaf2dfea8219)

lwasylow:
Nice I will pull the changes to bring my branch up to date.

jgebal:
I think I have solved the problem with 19c.
At least the tests are now running 🤞 

https://github.com/utPLSQL/utPLSQL/actions/runs/22830170857/job/66216998893
The problem was with how the `sqlnet.ora` was configured in the container. Similar to issue that existed in 21-XE.
https://github.com/gvenzl/oci-oracle-xe/issues/43

lwasylow:
Nice. 
For 21 and 23 we can switch also to slim faststart versions.

lwasylow:
@jgebal  

[...]

PR #1318, opened 2026-04-21
fix: prevent ora-21799 in ut_run()

martin-bach-oracle:
This PR addresses the _ORA-21799 duration not active_ error reported in https://forums.oracle.com/ords/apexds/post/ora-21779-duration-not-active-when-running-utplsql-on-23-26-4428

Initially I thought the error might be thrown during the transition to an autonomous transaction. The most suspicious spot in my investigation was crossing the autonomous boundary with object types in UT.RUN -> RUN_AUTONOMOUS (pragma autonomous_transaction) with a_reporter in out nocopy ut_reporter_base.

Turns out, that wasn't it. The RC appears earlier in the stack:

- The stack trace shows a failure at startup event: UT_RUNNER line 93 triggers gc_initialize (ut_runner.pkb (line 93))
- Event manager dispatches listener (ut_event_manager.pkb (line 70))
- Reporter handler does treat(a_event_item as ut_run) on initialize (ut_reporter_base.tpb (line 171))

For gc_initialize, a_event_item is passed as NULL (no event object). My suspicion is that the ORA-21779 occurs on object-duration/type-cast handling at event dispatch time.

I tried passing a real object in UT_RUNNER line 93:
```
ut_event_manager.trigger_event(ut_event_manager.gc_initialize, ut_run());
```

If the treat(NULL as ut_run) is the problem, that should fix the issue. It works on my machine. Kindly review and merge if applicable.

### DIFF
--- a/source/api/ut_runner.pkb
+++ b/source/api/ut_runner.pkb
@@ -90,7 +90,7 @@ create or replace package body ut_runner is
     end if;
     ut_event_manager.add_listener( ut_session_info() );
 
-    ut_event_manager.trigger_event(ut_event_manager.gc_initialize);
+    ut_event_manager.trigger_event(ut_event_manager.gc_initialize, ut_run());
     ut_event_manager.trigger_event(ut_event_manager.gc_debug, ut_run_info());
 
     if a_random_test_order_seed is not null then


lwasylow:
Hi @martin-bach-oracle
We have put a fix as part of https://github.com/utPLSQL/utPLSQL/pull/1310
Could you give it a shot using a develop source code branch?

jgebal:
Hi @martin-bach-oracle 

We also faced the same issue recently and a fix is already merged into develop branch.
There was a comment about it here:
https://github.com/utPLSQL/utPLSQL/issues/1309#issuecomment-4020289898

The fix that we have applied can be found here:
https://github.com/utPLSQL/utPLSQL/pull/1310/changes#diff-5725a92b6cb359b9213a9b3390c253f930d04d9072f89c7bec1c6717aba48e07R168


I will prepare new intermediate release of utPLSQL framework, java-api and utPLSQL-cli, to give them a small refresh.

martin-bach-oracle:
Thanks for following up.

I'm not 100% sure now but I _believe_ your fixes might still have thrown the error reported in the forum. Do you see any downside in changing line 93 in `ut_runner`?

jgebal:
I will investigate and let you know.

jgebal:
@martin-bach-oracle 
I checked the utPLSQL code for all references to `ut_event_manager.trigger_event`
There are multiple instances where we internaly call `ut_event_manager.trigger_event` without the second parameter.
Passing an empty `ut_run()` object is more like a hack/workaround then a solution as empty `ut_run()` object does not represent the context of current utPLSQL test run that is executed.
In some cases that context is not relevant and not needed.
This is why the second arcument of the procedure `ut_event_manager.trigger_event` is optional and defaults to null.

In general, optional NULL argument is correct functionality but the behavior of this peice of code was unstable.
`treat(a_event_item as ut_run)`

Because it "sometimes fails" and "sometimes works" I consider it to be an oracle bug.

The workaround done by @lwasylow maintains the behavior (we still have NULL argument) so I would consider this a better option for now.

If that doesn't work, then the case statement needs to be changed to prevent `treat(l_event_item as ut_run)` alltogether in cases when `l_event_item is NULL`.

After the fix, every run of our pipeline on 23.26 AI was successful so I am expecting the workaround to be stable.

@lwasylow  - do you have any insights into research you did and how you figured out that it's related to the fact that using the input argument variable as the problem?

martin-bach-oracle:
no worries, feel free to close the PR in that case

jgebal:
Thank you @martin-bach-oracle also for your feedback on oracle forum.

New releases should be available soon.

